PR #26, opened 2015-08-18
#24: package import is too verbose (closes #24)

### LINKED ISSUE #24: package import is too verbose
right now this is the standard way to import a package:
```
import LoadingSpinner from 'buildo-react-components/lib/loading-spinner';

require('buildo-react-component/lib/loading-spinner/style.css');
```

I think we should:
- export each lib directly from the main `index.js` so to achieve: `import {LoadingSpinner} from 'buildo-react-components';`

FrancescoCioria:
Issue #24

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,11 +1,20 @@
-import * as CookieBanner from './cookie-banner';
+import { CookieBanner, cookie } from './cookie-banner';
 import * as DatePicker from './datepicker';
+import Dropdown from './dropdown';
+import Flex from './flex';
 import InputLink from './input-link';
+import LoadingSpinner from './loading-spinner';
+import Popover from './popover';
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -1,11 +1,20 @@
...
 export default {
   CookieBanner,
-  DatePicker,
+  cookie,
+  ...DatePicker,
+  Dropdown,
+  ...Flex,
   InputLink,
+  LoadingSpinner,
+  Popover,
```
giogonzo: all better written as
`export * from 'src/other_module';`
or
`export { foo, bar } from 'src/other_module';`

also @francescogior had issues default exporting an object

giogonzo:
:+1:

PR #34, opened 2015-08-31
#33: add "events" object prop and pass it to first child (closes #33)

### LINKED ISSUE #33: add "events" object prop and pass it to first child

FrancescoCioria:
Issue #33

### DIFF
--- a/src/flex/FlexView.js
+++ b/src/flex/FlexView.js
@@ -3,58 +3,57 @@ import cx from 'classnames';
 import pick from 'lodash/object/pick';
 import omit from 'lodash/object/omit';
 
+const propTypes = {
+  children: React.PropTypes.node,
+  row: React.PropTypes.bool,
+  column: React.PropTypes.bool,
+  auto: React.PropTypes.bool,
+  vAlignContent: React.PropTypes.oneOf(['top', 'center', 'bottom']),
+  hAlignContent: React.PropTypes.oneOf(['left', 'center', 'right']),
+  marginLeft: React.PropTypes.oneOfType([
+    React.PropTypes.string,
+    React.PropTypes.number
+  ]),
+  marginTop: React.PropTypes.oneOfType([
+    React.PropTypes.string,
+    React.PropTypes.number
+  ]),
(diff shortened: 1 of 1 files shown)


src/flex/FlexView.js
```diff
@@ -43,6 +43,7 @@ export default React.createClass({
       React.PropTypes.number,
       React.PropTypes.string
     ]),
+    events: React.PropTypes.object,
```
giogonzo: not sure about `events` name: this is just an object of additional props, so maybe `props` would be more appropriate?

src/flex/FlexView.js
```diff
@@ -140,7 +142,11 @@ export default React.createClass({
   render() {
     const className = this.getClasses();
     const style = this.getStyle();
-    return <div className={className} style={style}>{this.props.children}</div>;
+    return (
+      <div className={className} style={style} {...this.props.events}>
```
giogonzo: maybe better to omit `className` and `style` from `props.events`? Duplicate props can cause unexpected behavior

src/flex/FlexView.js
FrancescoCioria: it is, but I wanted to be clear that it is intended to be used only for events as it will passed to first child which is a react `div`. I don't want it to be used for other purposes like `style`, `className` or `id`. Does react `div` accept non events props other than the three above?

src/flex/FlexView.js
FrancescoCioria: :+1:

FrancescoCioria:
rebased

src/flex/FlexView.js
giogonzo: All of the supported html attributes: https://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes
But there shouldn't be any "special" props other than maybe input props: https://facebook.github.io/react/docs/forms.html#interactive-props

Looking at other props, `id` is receiving exactly the same treatment as `events` props

Naming still confuses me, but maybe it's just me :P @gabro thoughts?

FrancescoCioria:
@gabro ping

PR #35, opened 2015-08-31
#32: components should accept 'id' (closes #32)

### LINKED ISSUE #32: components should accept 'id'
EDIT: for now only the components for which we have the source (`flex` and `loadingSpinner`)

FrancescoCioria:
Issue #32


gabro:
:+1:

PR #52, opened 2015-09-24
#49: [Popover] rename misleading type props "absolute" and "relative" (closes #49)

### LINKED ISSUE #49: [Popover] rename misleading type props "absolute" and "relative"
remove `absolute` and `relative` in favor of `attachToBody={true/false}`

/cc @giogonzo @francescogior

FrancescoCioria:
Issue #49

### DIFF
--- a/src/popover/Popover.js
+++ b/src/popover/Popover.js
@@ -7,7 +7,7 @@ const Popover = React.createClass({
     children: React.PropTypes.node.isRequired,
     popover: React.PropTypes.shape({
       content: React.PropTypes.node.isRequired,
-      type: React.PropTypes.oneOf(['absolute', 'relative']),
+      attachToBody: React.PropTypes.bool,
       position: React.PropTypes.oneOf(['top', 'bottom']),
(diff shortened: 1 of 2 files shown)


src/popover/Popover.js
```diff
@@ -223,7 +223,7 @@ const Popover = React.createClass({
   },
 
   isAbsolute() {
-    return this.getPopoverProps().type === 'absolute';
+    return this.getPopoverProps().attachToBody === true;
```
ascariandrea: You can simply return the value of `attachBody`

ascariandrea:
:+1:

src/popover/Popover.js
FrancescoCioria: I know, but I prefer it to as a boolean (`attachToBody` can be `undefined`)

ascariandrea: You can resolve by using `!!` operator, like `!!this.getPopoverProps().attachToBody`. It always returns a boolean ;)

PR #65, opened 2015-10-20
#61: implement TextOverflow component (closes #61)

### LINKED ISSUE #61: implement TextOverflow component
refer to https://github.com/buildo/labonline/issues/622 for specs (only the part about ellipsis and popover)

FrancescoCioria:
Issue #61


FrancescoCioria:
@giogonzo I renamed the branch so the previous PR has been deleted (it was set to wrong issue..)

giogonzo:
sorry I'm super late

will comment on code, but first, what's the envisioned strategy for: using `TextOverflow` with a custom `Popover`/`Tooltip` component (custom styled, etc)?

e.g. LOL would use `TextOverflow` but already has a custom `Tooltip` (which is LOL custom too, correct me @francescogior )

src/text-overflow/TextOverflow.js
```diff
@@ -0,0 +1,87 @@
...
+      anchor: React.PropTypes.oneOf(['left', 'center', 'right']),
+      onShow: React.PropTypes.func,
+      onHide: React.PropTypes.func,
+      dismissOnScroll: React.PropTypes.bool,
+      className: React.PropTypes.string,
+      id: React.PropTypes.string,
+      maxWidth: React.PropTypes.oneOfType([
+        React.PropTypes.number,
+        React.PropTypes.string
+      ])
```
giogonzo: are these all the popover props repeated here?

src/text-overflow/TextOverflow.js
```diff
@@ -0,0 +1,87 @@
...
+    props.style = {
+      width: '100%',
+      ...props.style
+    };
+
+    const text = this.getTextSpan();
+    if (isOverflowing) {
+      return <Popover { ...props }>{text}</Popover>;
+    } else {
+      return <div { ...props }>{text}</div>;
```
giogonzo: is the double wrapping required? `<div><span>text</span></div>`

FrancescoCioria: yes, because `Popover` adds a `div` and (I guess) we want the layout structure to be the same in both cases.

src/text-overflow/TextOverflow.js
FrancescoCioria: yes, this is stupid I agree.

FrancescoCioria:
My idea was to simply pass the `className` you're using to style that `Popover` wrapper (in this case `Tooltip`). If you think there might be some useful prop manipulation these wrapper do (other then `className`) we can add an optional prop to pass the component to use as `Popover` (of course it would default to the original `Popover`);

FrancescoCioria:
@giogonzo @francescogior ping

giogonzo:
@FrancescoCioria got it,

in general I favor *composing components* over *cascading styles*,

so I was thinking more of something like: accept a function `tooltipText => <AnyReactNode>` to customize the popover, or pass the custom popover as only child and have text as a string prop, but I'm not sure how better these solutions would be.

This is not a super issue in LOL terms, as we are using the basic `Tooltip` with no additional styling / additions there, but still.. what do you think?

giogonzo:
I don't see any better solution than:

```js
<TextOverflow label={possiblyLongString} /> // no popover customization

<TextOverflow label={possiblyLongString}>
  {(self) => ( // self: "ellipsified if needed" label, e.g. <div style={{ txtoverflow: ellipsis }}></div>
    <MyPopover content={label}>{self}</MyPopover>
  )}
</TextOverflow>
```

only bad thing is it expects the customization popover to have the same interface of our popover, so it should render the original content AND the popover on top of it itself.

FrancescoCioria:
@giogonzo edited as you suggested. should be safe to merge

giogonzo:
Ok, but trying this out on LONL first

giogonzo:
:+1:

PR #77, opened 2015-11-02
#74: toaster (notifications manager) (closes #74)

### LINKED ISSUE #74: toaster (notifications manager)
## requirements
discussed here: https://github.com/buildo/labonline/issues/730 + offline with pretty much everyone
final behavior: same as: http://igorprado.com/react-notification-system/ (but with a much nicer api!)
- display a list of dismissable notifications
- the list should be positioned at one of the corners top-left, top-right, etc (but we can start with top-right only), and stacking of toast notifications behave accordingly
  - e.g. for top-right, new notifications appear at the top of the stack and disappear at the bottom (FIFO) after a certain timeout.
- possible interaction is just "pausing" a specific item in the stack from disappearing automatically, how this interaction is initiated is delegated to the user / user component
  - when an item is paused, it will not disappear automatically, but other toasts in the stack behaves as they would so the entire stack can slide, including the "paused" toast
- items should be placed on top of everything and not interfer with the rest of the layout

## specs
- `Toaster` should have a stateless interface accepting ~ a list of children (all the toasts to be rendered / animated)
- internally, it is stateful and handles a timeout and a state (paused true/false) for each toast
- it should be able to accept a container element (id?) to be used as the container to render in
- should apply arbitrary animations (OSX?) for sliding and enter/leave, but consider this should be easily customizable in the future

**api draft**

**EDIT: old. look at comment below**
```js
<Toaster attachTo="#dasd">
  {(activeOn: fn, activeOff: fn, dismiss: fn) => errors.map(e => (
    <MyToastComponent
      key={uniqueToastKey(i)}
      onHover={activeOn}
      onHoverOut={activeOff}
      onDismissClick={dismiss(uniqueToastKey(i))}
    />
  )}
</Toaster>
```

FrancescoCioria:
Issue #74

### DIFF
--- a/examples/components/Toaster.js
+++ b/examples/components/Toaster.js
@@ -1,0 +1,42 @@
+import React from 'react';
+import Toaster from '../../src/toaster/Toaster';
+
+const Example = React.createClass({
+
+  propTypes: {},
+
+  getInitialState() {
+    return {
+      initialized: false
+    };
+  },
+
+  componentDidMount() {
+    this.setState({ initialized: true });
+  },
+
+  getTemplate() {
+
+    const toaster = (
+      <Toaster attachTo='toaster' className='hello' style={{backgroundColor: 'blue'}}>
+        <div style={{backgroundColor: 'yellow', height: 50, width: 200, marginTop: 10, marginRight: 20}} />
+        <div style={{backgroundColor: 'yellow', height: 50, width: 200, marginTop: 10, marginRight: 20}} />
+        <div style={{backgroundColor: 'yellow', height: 50, width: 200, marginTop: 10, marginRight: 20}} />
+        <div style={{backgroundColor: 'yellow', height: 50, width: 200, marginTop: 10, marginRight: 20}} />
+      </Toaster>
+    );
+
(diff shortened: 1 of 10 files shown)


FrancescoCioria:
@giogonzo this is the implementation of the first one of the two approaches I explained in the [issue](https://github.com/buildo/react-components/issues/74).

FrancescoCioria:
@giogonzo forget what I said. I refactored `Toaster` to use `ReactTransitionGroup` and implemented `ToastWrapper` component to handle inline styles.
I'll update `Toaster` API specs in the associated issue

FrancescoCioria:
@giogonzo after feedback from @ascariandrea I updated `Toaster` to support also inplace render:
`attachTo` is no longer required and if missing `Toaster` renders its content as a normal component by returning it in the `render` method.

FrancescoCioria:
Test on LOL went fine so I'm merging this. (I've already opened two issues about minor defects)

FrancescoCioria:
Ehm... Merging after a rebase... Which I can't do on smartphone :D

PR #92, opened 2015-11-11
#90: Add wrap prop to FlexView (closes #90)

### LINKED ISSUE #90: Add wrap prop to FlexView
## Requirements
- `FlexView` should optionally let content wrap in lines
  - to avoid this image
  - and let this happen 
image

## Specs
- `FlexView` should have an optional `wrap` prop

Prop | Type | Required | Default
---|---|---|---
wrap | bool | no | false

francescogior:
Issue #90

### DIFF
--- a/src/flex/FlexView.js
+++ b/src/flex/FlexView.js
@@ -46,6 +46,7 @@ const propTypes = {
     React.PropTypes.number,
     React.PropTypes.string
   ]),
+  wrap: React.PropTypes.bool,
   className: React.PropTypes.string,
   style: React.PropTypes.object
 };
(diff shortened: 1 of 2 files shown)


src/flex/FlexView.js
```diff
@@ -149,7 +150,8 @@ export default React.createClass({
   getClasses() {
     const direction = this.props.column ? 'flex-column' : 'flex-row';
     const contentAlignment = this.getContentAlignmentClasses();
-    return cx('react-flex-view', direction, contentAlignment, this.props.className);
+    const wrap = this.props.wrap && 'flex-wrap';
```
FrancescoCioria: I think this adds a class `"false"` it `this.props.wrap` is `false`/`undefined`. have you checked it?

FrancescoCioria: francescogior told me offline he verified on `classnames` repo that any falsy value is ignored.

francescogior: for reference https://github.com/JedWatson/classnames#usage

PR #102, opened 2015-11-12
#101: delete semantic-useless Modal component (closes #101)

### LINKED ISSUE #101: delete semantic-useless Modal component
`Modal` is a semantic wrapper component which we never used. Now that we made `ModalManager` we'd better remove it completely as it may seem connected with it

FrancescoCioria:
Issue #101


giogonzo:
:+1:

PR #106, opened 2015-11-13
#105: [ModalManager] remove BackgroundDimmer (closes #105)

### LINKED ISSUE #105: [ModalManager] remove BackgroundDimmer
## requirements
we should let the user decide if he needs it. This way he can also handle the click outside directly inside the modal which is the wanted behavior (dismiss modal action si dispatched form a single module instead of two).
the only way to this is accept children instead of modals array with `ReactClass`es.
The usual implementation already requires the user to have `MyProjectModalManager` where he creates the `modals` array -> It shouldn't be too hard providing `ModalManager` with already the right `ReactElement`

## specs
- remove `modals` prop
- remove `data` prop
- accept element as `children`. only one element, no array!
- use element `key` instead of `activeModal`

FrancescoCioria:
Issue #105


examples/components/Modal.js
```diff
@@ -1,39 +0,0 @@
-import React from 'react';
```
FrancescoCioria: forgot to delete this one in previous PR

examples/examples.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
```
FrancescoCioria: same here... I should have removed this in previous PR

giogonzo:
looks very good to me :+1:

PR #116, opened 2015-11-25
#115: [modal] add basic Modal wrapper component (closes #115)

### LINKED ISSUE #115: [modal] add basic Modal wrapper component
## requirements
to be used with `ModalManager`. should speed up development (or simply help as example) but at the same time be very basic so not to require modifications. It's a low-level component: in theory every modal should be able to use it out of the box.

## specs
- wrap content with `BackgroundDimmer`
- should accept optional `onDismiss` function -> if it exists:
  - render icon close in top right corner
  - close on click outside

FrancescoCioria:
Issue #115


giogonzo:
looks vvgood

PR #119, opened 2015-11-26
#112: [Popover] wrong positioning when hidden (closes #112)

### LINKED ISSUE #112: [Popover] wrong positioning when hidden
## requirements
the popover should be positioned correctly also when `visibility: hidden` so to avoid corruption of window size

## specs
when the poopver is not open:
- render a hidden version of the popover (`visibility: hidden`)
- wrap it in a div with:
  - absolute positioning
  - no size (no width and no height)
  - overflow hidden
this way the hidden popover will not affect the app layout in anyway

EDIT 26 Nov

FrancescoCioria:
Issue #112

### DIFF
--- a/examples/components/Popover.js
+++ b/examples/components/Popover.js
@@ -33,7 +33,7 @@ const Example = React.createClass({
     return {
       content,
       position: 'bottom',
-      anchor: 'end',
+      anchor: 'center',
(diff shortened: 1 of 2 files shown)


giogonzo:
:open_mouth:

PR #120, opened 2015-11-26
#109: [Popover] Popover.popover.dismissOnClickOutside is too fragile (closes #109)

### LINKED ISSUE #109: [Popover] Popover.popover.dismissOnClickOutside is too fragile
See for instance: https://github.com/buildo/hailadoc/pull/1400#discussion_r45472408

Basically, as you're listenting for click on `window`, anyone else could intercept that event before the dismiss handler..

We could use a transparent overlay (same as in Dimmer), but this will prevent actually clicking on other things on the page (first click closes the popover, second click will actually work)

I know you both have opinions about this @FrancescoCioria @francescogior

FrancescoCioria:
Issue #109

### DIFF
--- a/examples/components/Popover.js
+++ b/examples/components/Popover.js
@@ -36,7 +36,7 @@ const Example = React.createClass({
       anchor: 'center',
       attachToBody: true,
       dismissOnClickOutside: true,
-      dismissOnScroll: true,
+      dismissOnScroll: false,
       event: 'click',
       offsetX: 0,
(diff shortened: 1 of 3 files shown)


FrancescoCioria:
@giogonzo  maybe we should discuss this a bit more before merging. it works as expected (the transparent layer blocks the interaction with other elements) but I'm not sure we want this..

giogonzo:
commenting on issue about this

PR #123, opened 2015-11-26
#113: [FlexView] consistency between grow, shrink and flexBasis names (closes #113)

### LINKED ISSUE #113: [FlexView] consistency between grow, shrink and flexBasis names
## requirements
`flex-grow`, `flex-shrink` and `flex-basis` should have consistent names in `FlexView` API

## specs
- as it's pretty obvious we're talking about `flex` stuff, I'd change `flexBasis` to simply `basis`... but, as we're intensively using this prop, we should keep accepting `flexBasis` for now, so not to break projects....

FrancescoCioria:
Issue #113

### DIFF
--- a/examples/components/FlexView.js
+++ b/examples/components/FlexView.js
@@ -25,10 +25,12 @@ const Example = React.createClass({
               style={{backgroundColor: 'red', height: 30, width: 30}}
               marginTop='auto'
               marginBottom='auto'
+              basis={200}
               />
             <FlexView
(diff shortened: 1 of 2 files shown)


giogonzo:
Not a major issue, but we can easily `warn` for deprecated props.. what do you think?

FrancescoCioria:
If you think this could become a common practice  I could make a utility mixin so to implement it quickly when we need to

PR #136, opened 2015-12-30
#135: raw examples file to work with playground (closes #135)

### LINKED ISSUE #135: raw examples file to work with playground
## requirements
we need each example to be imported as plain js text (without `import`/`export`) so to work with `component-playground` from our future wonderful showroom

## specs
- remove `import`/`export` from each example
- refactor `examples.js` to work using `component-playground`

FrancescoCioria:
Issue #135


FrancescoCioria:
discussed offline with @giogonzo

PR #138, opened 2015-12-30
#137: use buildo/react-kitchen-sink to render examples (closes #137)

### LINKED ISSUE #137: use buildo/react-kitchen-sink to render examples

FrancescoCioria:
Issue #137


FrancescoCioria:
- big refactor of examples:
  - use `react-kitchen-sink` (actually now it's a `react-component` but it's not exported)
  - handle sections
  - handle multiple examples for single component
  - split multiple examples in multiple files

test plan:
- No source code was touched.
- the new component `kitchen-sink` is not exported from `/src` (isolated from outside)

PR #141, opened 2016-01-01
#140: [ModalManager] context wrapper breaks leave transition (closes #140)

### LINKED ISSUE #140: [ModalManager] context wrapper breaks leave transition
## bug report
the current way of passing context to separate react node breaks the leave transition

FrancescoCioria:
Issue #140

### DIFF
--- a/src/modal-manager/ModalManager.js
+++ b/src/modal-manager/ModalManager.js
@@ -24,6 +24,20 @@ const ModalManager = React.createClass({
     };
   },
 
+  componentWillMount() {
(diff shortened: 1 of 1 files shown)


giogonzo:
nice! so this also preserves the context passing.. looks like a win-win

If `ContextWrapper` ever gets in the way again feel free to drop support for it - as already discussed we're not making use of it anywhere now

(btw we should probably give `ModalManager` one more spin on LOL after this change, this might solve other mysteriously related issues we had)

PR #142, opened 2016-01-02
#73: remove outside libraries from package (closes #73)

### LINKED ISSUE #73: remove outside libraries from package
I think it's more painful then useful to export external libraries. Right now we're exporting these:
- `rc-datepicker`
- `react-input-link`
- `react-autosize-textarea`
- `react-cookie-banner`

We should import them in our projects directly from their npm packages so to remove one step of synchronization

( right now I'm having problem with `npm shrinkwrap` on hailadoc because of a dffference in `react-input-link` version between global and `buildo-react-components` ... )

FrancescoCioria:
Issue #73


giogonzo:
Looks good, only we should check how this affects HAIL and LOL (where we're depending on this module from master) and fix if needed

FrancescoCioria:
LOL doesn't depend on any of them through b-r-c.
Hailadoc did depend on `react-components` for `autosize-textarea` but I fixed it with this PR https://github.com/buildo/hailadoc/pull/1599

PR #144, opened 2016-01-04
#143: [TextOverflow] verify if setTimeout is needed (closes #143)

### LINKED ISSUE #143: [TextOverflow] verify if setTimeout is needed
## fix
verify if we actually need this `setTimeout`: https://github.com/buildo/react-components/blob/master/src/text-overflow/TextOverflow.js#L53

ascariandrea:
Issue #143

The `setTimeout` seemed useless and it was the cause of some exceptions when the component was forced to render.
Consider `setTimeout()` with no delay put its callback in the next execution queue, the forced rendered `TextOverflow` threw exception if the component has unmounted in the meantime.

screen shot 2016-01-04 at 12 37 51

The attachment shows the lifecycle of `TextOverflow` components.

### DIFF
--- a/src/text-overflow/TextOverflow.js
+++ b/src/text-overflow/TextOverflow.js
@@ -50,7 +50,7 @@ const TextOverflow = React.createClass({
       if(text.offsetWidth < text.scrollWidth){
         this.setState({ isOverflowing: true });
       }
-      setTimeout(this.logWarnings);
+      this.logWarnings();
     }
   },
 


FrancescoCioria:
great!

giogonzo:
I guess the `setTimeout` rationale was to wait an arbitrary amount of time before doing the warning check? If yes this also makes the `logWarnings` useless? @FrancescoCioria

FrancescoCioria:
@giogonzo I thought the `setTimeout` was there to log warning even if in the case of `text.offsetWidth < text.scrollWidth` (where we do a `setState`), but looking at the code it looked like, after the `setState`, it would start another cycle with another `componentDidUpdate` but I tested it and it doesn't look like that's the way it works!

So yeah, we should have tested it better...
I'm gonna do another PR with the right code.

PR #155, opened 2016-01-12
#154: Remove non-direct css inheritance panel->panel-header' (closes #154)

### LINKED ISSUE #154: Remove non-direct css inheritance panel->panel-header'
## Requirementes
- add `>` between `.panel` and `.panel-header` selectors so that nested panels don't get affected

francescogior:
Issue #154

### DIFF
--- a/src/Panel/panel.scss
+++ b/src/Panel/panel.scss
@@ -29,7 +29,7 @@
     margin: 10px;
   }
 
-  &.collapsable .panel-header {
+  &.collapsable > .panel-header {
     padding-left: 0;
   }
 


FrancescoCioria:
what about this part: https://github.com/buildo/react-components/blob/react_components-154-remove_nondirect_css_inheritance/src/Panel/panel.scss#L36-L60

francescogior:
I think that part is OK!
Anyway I prefer to make tiny and fine grained fixes to avoid at most to break stuff

FrancescoCioria:
mmm, I disagree: I checked and `panel-soft-loader` is indeed a sibling of the other ones you fixed, so there should be no problem in fixing that too. I think we should treat all of them the same. If for some absurd reason we break something we can always come here and fix it again :)

FrancescoCioria:
my comment above wasn't very relevant to this particular PR (it's more about missing sass guidelines)
so I'll merge this!

francescogior:
As discussed offline, the purpose of this commit was to avoid that a `.panel > .panel-header` nested into a `.panel.collapsable > .panel-header` inherited unwanted css properties

PR #172, opened 2016-01-15
#171: use buildo/eslint-config (closes #171)

### LINKED ISSUE #171: use buildo/eslint-config
## requirements
current config of `eslint` is missing many useful rules (like `no-undef`...)

## specs
let's use `buildo/eslint-config` even if it's not complete yet

FrancescoCioria:
Issue #171


FrancescoCioria:
@giogonzo I removed `/src/revenge` (I forgot to delete it in the past...) I checked on labonline and it's not depending on that folder so it's safe to remove it

FrancescoCioria:
the only logic I added is here: https://github.com/buildo/react-components/blob/171-use_buildoeslintconfig/src/utils/log.js
I'm using this function to:
- have `process.env.NODE_ENV !== 'production'` in a single place
- bypass `eslint` `no-console` rule for warnings

src/utils/log.js
```diff
@@ -0,0 +1,10 @@
+function warn(content) {
+  if (process.env.NODE_ENV !== 'production') {
+    const message = (typeof content === 'function') ? content() : content;
+    if (message) {
+      console.warn(message); // eslint-disable-line no-console
+    }
+  }
+}
+
+export default { warn }
```
giogonzo: NOTE: IMO this would be more readable as `export function warn( // ...`

instead of declaration and export (using default?) in separate places

giogonzo:
:+1: looks good

Note that this https://github.com/eslint/eslint/pull/4932 is now merged, but it could be the scope of another PR

src/utils/log.js
FrancescoCioria: Yeah, I thought I had other logs apart from `warn`, so I structured it to have `export default { warn, error, log }`. Not sure if we'll ever have them..

PR #179, opened 2016-01-18
#178: Button should accept custom TextOveflow (closes #178)

### LINKED ISSUE #178: Button should accept custom TextOveflow
## requirements
We should be able to use a custom wrapper of `TextOverflow` so to have project custom defaults (ex: `attachToBody: true`)

## specs
add prop `textOverflow` -> expect a class

FrancescoCioria:
Issue #178


FrancescoCioria:
@francescogior has already tested it on labonline

giogonzo:
:+1:

PR #181, opened 2016-01-19
#180: remove TooltippedIcon (closes #180)

### LINKED ISSUE #180: remove TooltippedIcon
## requirements
we moved it back to Labonline. we shouldn't have a duplicate (may cause confusion)

## specs
- delete `TooltippedIcon`
- fix `Panel` and `PanelHeader` api
- remove `PanelMenu`

FrancescoCioria:
Issue #180


src/Panel/Panel.js
```diff
@@ -21,9 +20,7 @@ export const Props = {
     content: t.maybe(t.ReactNode),
     title: t.maybe(t.ReactNode),
     hideTitleWhenExpanded: t.maybe(t.Bool),
-    menu: t.maybe(t.struct({
-      ...panelMenuProps
-    }))
+    menu: t.maybe(t.ReactNode)
```
giogonzo: can you elaborate a bit on this change?

src/Icon/index.js
```diff
@@ -1,2 +1,2 @@
-export Icon from './Icon';
-export TooltippedIcon from './TooltippedIcon';
\ No newline at end of file
+import Icon from './Icon';
+export default Icon;
```
giogonzo: this could be a one liner? (not sure)

FrancescoCioria: yes, I didn't know it. (every other component is exported like this as well..)

src/Panel/Panel.js
FrancescoCioria: `Panel` was built to be general and so was `PanelHeader` who at the beginning only had `title` and `content` both accepted as `t.ReactNode`
By adding more and more features we added the `menu` prop too, but this time instead of adding it as `t.ReactNode` (which is general and consistent with the other two header parts) we added it as an object which was directly passed to `PanelMenu`.

What I did was accepting a general `ReactNode` instead of some particular props for `PanelMenu` so it's up to the user to pass the rendered component they wish to use. (go https://github.omnilab.our.buildo.io/buildo/labonline/pull/1433 to see how i handled this in LOL)

giogonzo: got it, sounds better and reasonable

giogonzo:
:+1:

PR #183, opened 2016-01-19
#182: Rewrite weird `export default { ... }`s (closes #182)

### LINKED ISSUE #182: Rewrite weird `export default { ... }`s
## requirements
- clean up

## specs
```js
const a = something;
const b = somethingElse;
export default { a, b };
```
is better written as
```js
export const a = something;
export const b = somethingElse;
```
It shouldn't break anything considering the way we import

giogonzo:
Issue #182

## test plan
imports on different projects shouldn't break (tested on LOL only)

### DIFF
--- a/src/flex/index.js
+++ b/src/flex/index.js
@@ -1,4 +1,2 @@
-import FlexView from './FlexView';
-import FlexCell from './FlexCell';
-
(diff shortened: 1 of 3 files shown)


FrancescoCioria:
:+1:

PR #185, opened 2016-01-19
#184: implement Toggle (closes #184)

### LINKED ISSUE #184: implement Toggle
## requirements
- simple toggle done exclusively with `css`.
- props:
  - `value: t.Boolean`
  - `onChange: t.Function`

NOTE: I chose `value` and `onChange` as they're the standard for form element

FrancescoCioria:
Issue #184

**result**
image
image

### DIFF
--- a/src/toggle/Toggle.js
+++ b/src/toggle/Toggle.js
@@ -45,14 +45,13 @@ export default class Toggle extends React.Component {
 
   getHalfSize(size) {
     if (t.String.is(size)) {
-      const numberMatch = (/\d+/).exec(size);
       const unitMatch = (/[a-z]+/).exec(size);
 
-      const number = numberMatch ? parseInt(numberMatch[0], 10) : '';
+      const number = parseFloat(size, 10);
       const unit = unitMatch ? unitMatch[0] : '';
       return `${number / 2}${unit}`;
     } else {
-      return size / 2
+      return size / 2;
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/toggle/Toggle.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  getHalfSize(size) {
+    if (t.String.is(size)) {
+      const numberMatch = (/\d+/).exec(size);
+      const unitMatch = (/[a-z]+/).exec(size);
+
+      const number = numberMatch ? parseInt(numberMatch[0], 10) : '';
+      const unit = unitMatch ? unitMatch[0] : '';
+      return `${number / 2}${unit}`;
```
giogonzo: This is scary and not safe for corner cases like '13.5px' or 'px'. No easy way to get the computed size instead? Otherwise you can rely on the scary but spec'd feature of `parseInt/parseFloat` of ignoring non-digits in a string when parsing

giogonzo:
Looks good except for comment

FrancescoCioria:
@giogonzo is it better now?

giogonzo:
better, but I'd add two more checks:

```js
const unitMatch = (/[a-z]+$/).exec(size); // only match characters at the end
const number = parseFloat(size, 10);
const unit = unitMatch ? unitMatch[0] : '';
if (isFinite(number)) { // we can still get NaN from parseFloat
  return `${number / 2}${unit}`;
} else {
  warn('Invalid size');
  return 0;
}
```

giogonzo:
:+1:

PR #187, opened 2016-01-19
#186: improve ConfirmationInput (closes #186)

### LINKED ISSUE #186: improve ConfirmationInput
refer to https://github.omnilab.our.buildo.io/buildo/labonline/issues/1243
## requirements
- `Applied` label should be replaced by `Clear` action
  - show only if **NOT** on focus and **NOT** empty

FrancescoCioria:
Issue #186

### DIFF
--- a/examples/components/confirmation-input/default.example
+++ b/examples/components/confirmation-input/default.example
@@ -7,13 +7,15 @@ class Example extends React.Component {
 
   onConfirm = (value) => this.setState({ value })
 
+  onClear = () => this.setState({ value: '' })
+
   render() {
-    const { state: { value }, onConfirm } = this;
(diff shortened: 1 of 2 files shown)


FrancescoCioria:
@giogonzo NOTE: `onMouseLeave` is never called if its element is unmounted so I had to manually call it in the two cases `templateConfirm` is completely unmounted: after `onClear` and after `onConfirm` if the value is falsy (`''`)

FrancescoCioria:
@giogonzo ping

giogonzo:
as already discussed offline, it's complex and I don't like it (even though we couldn't come up with any cleaner solution at the moment)

I hope you tested it thoroughly ;)

PR #189, opened 2016-01-20
#134: [Toaster] div containers should be transparent to events (closes #134)

### LINKED ISSUE #134: [Toaster] div containers should be transparent to events
## requirements
`Toaster` should let every event pass below it

## specs
as its children have `position: absolute` we could simply make `Toaster` sizeless by applying `width: 0, height: 0`

FrancescoCioria:
Issue #134


francescogior:
After an accurate review, I'd say it's OK :fo

PR #191, opened 2016-01-20
#190: Badge shouldn't fill container (closes #190)

### LINKED ISSUE #190: Badge shouldn't fill container
## requirements
`Badge` should be an `inline-block` but it should keep the benefits of being a `FlexView` (children centered)

## specs
add `div` wrapper with `display: inline-block`

FrancescoCioria:
Issue #190


francescogior:
As you showed me offline, :+1:  :factory:

PR #203, opened 2016-01-21
#202: Safer way to get computed style of dom nodes, better warnings (closes #202)

### LINKED ISSUE #202: Safer way to get computed style of dom nodes, better warnings
`node.style` returns only the inline style. We should use `window.getComputedStyle(node)`  (https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) so to get style from CSS too
## Specs
- use `window.getComputedStyle(DOMNode)` instead of `DOMNode.style`
- log also the affected `DOMNode`

francescogior:
Issue #202


FrancescoCioria:
:8ball:

PR #209, opened 2016-01-21
#208: 'Menu should not have hover effect on disabled options' (closes #208)

### LINKED ISSUE #208: 'Menu should not have hover effect on disabled options'
That was ugly

francescogior:
Issue #208 

Much better


FrancescoCioria:
:100:

PR #216, opened 2016-01-22
#215: use `React.findDOMNode` anywhere (closes #215)

### LINKED ISSUE #215: use `React.findDOMNode` anywhere
switching to ES6 classes apparently caused some unexpected breaking changes.
We should replace `this.refs.{my_ref}.getDOMNode()` with `React.findDOMNode(this.refs.{my_ref})` to avoid bugs and increase compatibility with `React 0.14` (soon or later we'll have to make the jump :) )

FrancescoCioria:
Issue #215


giogonzo:
:+1:

Issue #204, opened 2016-01-21
standardize code


FrancescoCioria:
## requirements
code on this repo should follow stricter rule so to avoid differences between components (ex: new frontends are not used to `React.createClass({})`)

## sub-issues

- [x] use ES6 classes #207 
- [x] use tcomb + tcomb-react for prop evaluation
- [x] update buildo/eslint-config #205 
- [x] import b-r-c components directly #213
- [x] every folder should be in `kebab-case` #188 **[FrancescoCioria]**

## without issue
- [ ] ~~add docs to every component~~
- [ ] ~~each component should export its `props`~~
- [ ] ~~SASS files should follow our naming convention~~
- [ ] ~~`style`, `className`, `id` to every component~~

FrancescoCioria:
closing in favor of #490

PR #218, opened 2016-01-22
#217: use @props + t-comb for prop evaluation (closes #217)

### LINKED ISSUE #217: use @props + t-comb for prop evaluation
We're using `tcomb` + `tcomb-react` as the standard validators in our projects. We should do the same here :)

FrancescoCioria:
Issue #217

### DIFF
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -2,6 +2,7 @@ import React from 'react';
 import Select from 'react-select';
 import omit from 'lodash/object/omit';
 import cx from 'classnames';
+import { props, t } from '../utils';
 import { warn } from '../utils/log';
 
 const themes = {
@@ -9,21 +10,15 @@ const themes = {
 };
 
 const PropTypes = {
-  children: React.PropTypes.oneOfType([
-    React.PropTypes.node,
-    React.PropTypes.element,
-    React.PropTypes.array
-  ]),
-  theme: React.PropTypes.oneOf(['semantic']),
-  valueLink: React.PropTypes.shape({
-    value: React.PropTypes.string,
-    requestChange: React.PropTypes.func
-  })
+  children: t.ReactNode,
+  theme: t.maybe(t.enums.of(['semantic'])),
+  valueLink: t.maybe(t.struct({
+    value: t.maybe(t.String),
+    requestChange: t.Function
+  }))
 };
-
+@props(PropTypes, { strict: false })
 export default class Dropdown extends React.Component {
-
-  static propTypes = PropTypes
 
   getChildren = () => [].concat(this.props.children || [])
 
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -10,7 +10,7 @@ const themes = {
 };
 
 const PropTypes = {
-  children: t.ReactNode,
+  children: t.maybe(t.ReactNode),
   theme: t.maybe(t.enums.of(['semantic'])),
   valueLink: t.maybe(t.struct({
     value: t.maybe(t.String),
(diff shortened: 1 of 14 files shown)


src/dropdown/Dropdown.js
```diff
@@ -9,22 +10,16 @@ const themes = {
...
-    React.PropTypes.node,
-    React.PropTypes.element,
-    React.PropTypes.array
-  ]),
-  theme: React.PropTypes.oneOf(['semantic']),
-  valueLink: React.PropTypes.shape({
-    value: React.PropTypes.string,
-    requestChange: React.PropTypes.func
-  })
+  children: t.ReactNode,
```
giogonzo: just noting that: `children: oneOfType()` was not `.isRequired` before, now it is (probably better anyway)

FrancescoCioria: Mmm no it's a mistake thanks (you can use `options` as this is a wrapper of `react-select`)

src/dropdown/Dropdown.js
```diff
@@ -9,22 +10,16 @@ const themes = {
...
+  theme: t.maybe(t.enums.of(['semantic'])),
+  valueLink: t.maybe(t.struct({
+    value: t.maybe(t.String),
+    requestChange: t.Function
+  }))
 };
-
+@props(PropTypes, { strict: false })
```
giogonzo: we should add something like "you can pass any other prop and react-select it will be passed to react-select as-is"?

giogonzo:
looks great except for comment :+1: 

one more general comment/question is: how many different situations/strategies we have when dealing with "non strict" prop typing? I mean:
- "general props" case (style, className, id, ?): they are listed in `@props` with no docs, as it is added automatically (no `strict: false` needed)
- "wrapped component api" case: we wrap a first or third party component in another one, but we want to keep the underlying component api accessible in a flat fashion (`strict: false` is needed). this is the case of e.g. `DropDown` and also `TransitionWrapper` I guess (it can wrap any component)
- something else I'm missing?

src/dropdown/Dropdown.js
FrancescoCioria: yes, definitely. I was talking with @gabro about this issue two days ago: are `docgen` should identify `{ strict: false}` and also parse eventual comment describing **why** props are not strict.
PS: I'm going to add this comment when working on the *"add inline docs"* part of #204

FrancescoCioria:
> "general props" case

I think we should find a better way to handle these three standard props (`style`, `className`, `id`)
First of all, currently, not every component accepts all of them and I think they should, do you agree?

also:
> they are listed in @props with no docs, as it is added automatically (no strict: false needed)

this is not true for the new `revenge-docgen` made by @gabro (but I'll soon open an issue when I'll be working on docs here ;) )

> something else I'm missing?

there may be some cases (`FlexView`) where you want to let user set event listeners on your component (`onClick`, `onMouseEnter`....)

giogonzo:
> First of all, currently, not every component accepts all of them and I think they should, do you agree?

Yes

> this is not true for the new revenge-docgen made by @gabro (but I'll soon open an issue when I'll be working on docs here ;) )

got it, makes sense

> there may be some cases (FlexView) where you want to let users set event listeners on your component (onClick, onMouseEnter....)
PS: this might be considered a "wrapped component api" for div

Yup, although I'd also value keeping apis small and the "do one single thing right" approach: event handlers can probably always be placed in a wrapper or content of a `FlexView`. This same reasoning actually also applies to other general props (`id`, `className`, `style`), it's a just a trade-off between simplicity of implementation and simplicity of usage
> I think we should find a better way to handle these three standard props (style, className, id)

If and when we have a standard way of dealing with "other props" we can apply it to `FlexView` and other components

PR #232, opened 2016-01-27
#231: FlexView should set minHeight, minWidth to 0 by default (closes #231)

### LINKED ISSUE #231: FlexView should set minHeight, minWidth to 0 by default
## bug-report
for some reasons, on some browsers (on some browser VERSIONS), `flex` behaves in a weird way (and IMO wrong!)

## specs
`min-height: 0` and `min-width: 0` should already be the defaults -> we're not risking anything by enforcing it (yes, of course...)

FrancescoCioria:
Issue #231

### DIFF
--- a/src/flex/flexView.scss
+++ b/src/flex/flexView.scss
@@ -2,6 +2,9 @@
 
 .react-flex-view {
   box-sizing: 'border-box';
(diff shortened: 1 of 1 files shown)


giogonzo:
:goberserk:

PR #241, opened 2016-01-29
#240: ScrollView stop propagation should work also with visible scrollbars (closes #240)

### LINKED ISSUE #240: ScrollView stop propagation should work also with visible scrollbars

FrancescoCioria:
Issue #240


giogonzo:
:+1:

PR #242, opened 2016-01-29
#239: add ScrollView.onScroll prop (closes #239)

### LINKED ISSUE #239: add ScrollView.onScroll prop
## requirements
- after https://github.com/buildo/react-components/pull/238, `onScroll` prop is not honored by `ScrollView`, so there's no way for users to know e.g. the current `scrollTop` for the scrollview container.
 We should call `onScroll` if we receive it as prop

## specs
- to be discussed, but probably call `onScroll` on every event handled by `ScrollView` internally (`onScroll`, `onWheel`, ..?)
- TODO

/cc @FrancescoCioria

nerder:
Issue #239

## Test Plan
-  Go on the example page [ScrollView](http://localhost:8080/#/componentId=ScrollView&sectionId=section-2 )
- Try to scroll down and you and will see buttons appear and disappear
- Try to scroll down with mobile touch event (using CDT) and you will see buttons appear and disappear
Note: onWheel as been commented because it have a wired side effect.

### DIFF
--- a/src/scroll/ScrollView.js
+++ b/src/scroll/ScrollView.js
@@ -30,6 +30,10 @@ const PropTypes = {
    * easing function used when scrolling with `scrollTo`
    */
   easing: t.maybe(t.enums.of(Object.keys(easing))),
+  /**
+   *  onScroll function that handle onScroll events
+   */
+  onScroll: t.maybe(t.Function),
   style: t.maybe(t.Object)
 };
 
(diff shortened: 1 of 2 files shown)


src/scroll/ScrollView.js
```diff
@@ -40,19 +44,20 @@ export default class ScrollView extends React.Component {
...
 
   getScrollView = () => React.findDOMNode(this.refs.scrollView);
 
   getEventListeners = () => {
-    return !this.props.scrollPropagation ? {
-      onScroll: this.stopScrollPropagation,
-      onWheel: this.stopScrollPropagation,
+    return {
+      onScroll: this._onScroll,
+      //onWheel: this._onScroll,
```
FrancescoCioria: this will break `ScrollView` as `onScroll` currently doesn't work as expected

FrancescoCioria:
nice!

PR #249, opened 2016-02-03
#247: tableWrapper.scss should be easy to customize (closes #247)

### LINKED ISSUE #247: tableWrapper.scss should be easy to customize
## requirements
- easily customizable through sass variables
- shouldn't be imported by default (we don't want it twice in our code!)

## specs
- better names for variables
- each variable should be overridable (add `!default` at the end of each declaration)
- don't import variables from `b-r-c`

PS: look at style from labonline too as they have slightly diverged

FrancescoCioria:
Issue #247


giogonzo:
:+1:

PR #257, opened 2016-02-05
#256: Remove keypress listener and handlers from Table (closes #256)

### LINKED ISSUE #256: Remove keypress listener and handlers from Table
- in `Table` component keypress listeners and handlers should be removed
- they can be optionally handled by a specific wrapper

francescogior:
Issue #256

- in LOL the keyboard shortcuts should work as usual
- **NB**: do not remove `tabIndex` prop or `Shortcut` would stop to work


FrancescoCioria:
nice, thanks!

PR #260, opened 2016-02-08
#258: Meter (closes #258)

### LINKED ISSUE #258: Meter
## Requirements

- name: Meter
- there should be a bar
  - should be filled based on the current value (always LTR)
  - minimum value could be <> 0
  - can be colored based on different values
    - but always a single color
- there should be a label
  - positioned right of the bar
  - could be a numeric (currentValue/max) or percentage value [different style]
  - can be colored based on different values

## Specs

- `<Meter>` should have a stateless interface and do not accept any child
- it have to take `current` in order to show the result and the progress
- each part of the component can be customizable in style and behaviour
  - label format and coloring
  - color of the filling
- it can be splitted in different steps. Each step accept a range value and can have a custom style as well

## API

```
<Meter 
  value={Number}
  min={Number}
  max={Number}
  labelFormatter={Function}
  steps={Array}
  baseFillingColor={String}
  baseLabelColor={String}
/>
```

## Defaults

**min**:  `0`
**max**:  `100`
**labelFormatter**:
```
labelFormatter = (current, min, max) => {
  return `${computeResult(current, min, max)}%`;
};
```
**steps**: 
```
[{ 
  startValue: min,
  endValue: max,
  fillingColor: defaultFillingColor,
  labelColor: defaultLabelColor
}]
```

nerder:
Issue #258

## Test Plan

- run `npm start` and check the examples
- i provide examples for every use case
- Open console and you will see 2 warning, in order to cover from some bad behaviour
- unit tests in #275


src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
+import React from 'react';
+import { props, t, skinnable } from '../utils';
+import cx from 'classnames';
+import FlexView from '../flex/FlexView';
+import find from 'lodash/collection/find';
+import { warn } from '../utils/log';
+
+const Range = t.refinement(t.struct({
+  startValue: t.maybe(t.Number),
```
giogonzo: so we ended up making `startValue` optional/implicit? care to elaborate a little?

src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
...
+  endValue: t.Number,
+  fillingColor: t.String,
+  labelColor: t.String
+}), r => r.startValue < r.endValue);
```
giogonzo: protcombtip: always give custom types a name (last parameter of any combinator is an optional string `name`). It makes debugging less painful

src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
...
+
+const labelFormatter = (currentValue, startValue, endValue) =>(
```
giogonzo: proeslintip: always have the latest `buildo/eslint-config` installed.. it would have caught the missing space after `=>`

src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
...
+export default class Meter extends React.Component{
+
+  static defaultProps = {
+    minValue: 0,
+    maxValue: 100,
+    labelFormatter
+  };
+
+  validateSteps = (steps) => {
+    if(!steps){
```
giogonzo: `eslint keyword-spacing` should complain here

src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
...
+    } else {
+      return steps.reduce(( memo, curr, index, arr ) => {
+        if(index === 1){
+          memo.startValue = this.props.minValue;
+        }
+        curr.startValue = arr[index-1].endValue + 1; //scale to be defined
+        return memo;
+      });
+    }
+  }
```
giogonzo: about this validate steps function, just one thing to say for now: IMO it should be part of props validations, i.e. a `refinement` for the `steps` prop type

src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
...
+
+  getSteps = () => {
+    const {
+      steps,
+      minValue,
+      maxValue
+    } = this.props;
+
+    if(this.validateSteps(steps)){
```
giogonzo: if we know we have either `undefined` or a valid `steps` list as prop (see previous comments), than we can just `return this.props.steps || [{ ...defaultSingleStep }]` here

src/meter/Meter.js
```diff
@@ -0,0 +1,157 @@
...
+    const result = this.getResultFromCurrentValue({ currentValue, minValue, maxValue });
+    return {
+      wrapperStyle: {
+        width: result+'%',
+        height: '100%',
+        backgroundColor: step.fillingColor || '#ccc'
+      },
+      labelStyle: {
+        color: step.labelColor || '#000'
+      }
```
giogonzo: you should honor the default label color and default filling color props here, which I see are not part of the props api yet..

src/meter/Meter.js
nerder: @giogonzo i choose to make it implicit cause in most of the cases steps are contiguous, so they can be pre-computed. If the user want to force a new `startValue` it can be possibile as well.

src/meter/Meter.js
nerder: i'll fix this ASAP.

src/meter/Meter.js
nerder: I guess i'm the last version of eslint configuration installed, but i don't know why is not working correctly. To be verified.

src/meter/Meter.js
nerder: as the previous comment.

src/meter/Meter.js
nerder: do you mean that i have to let the user choose is own default?

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+  };
+
+  return noOverlappingRanges(rangeList);
+}, 'Ranges');
+
+const Props = t.refinement(t.struct({
+  /**
+   * This is the value provided as input.
+   */
+  current: t.Number,
```
FrancescoCioria: the standard way to pass values to this kind of react components is through `value` and (optionally) `defaultValue`

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+ * ### Renders a Progress Bar
+ */
+@skinnable()
+@props(Props)
+export default class Meter extends React.Component{
+
+  static defaultProps = {
+    min: 0,
+    max: 100,
+    defaultFillingColor: '#ccc',
```
FrancescoCioria: why do you need to set a default color? I would remove this and log a warning

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+    defaultLabelColor: '#000',
```
FrancescoCioria: IMO the default color for the label should be inherited (this is the standard for strings in HTML)

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+    };
+  }
+
+  getLocals(){
+    const {
+      className,
+      ...props
+    } = this.props;
+
+    const styles = this.computeStyle();
```
FrancescoCioria: `computeStyle(S)` ?

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+      labelStyle: styles.labelStyle
+    };
+  }
+
+  template({ className, style, labelStyle, ...locals }){
+    return (
+      <FlexView
+        className={className}
+        grow
+        row
```
FrancescoCioria: row is not a prop of `FlexView`. please remove (only `column` is a valid prop and by default is `false`)

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+        style={style}
+      >
+        <FlexView
+          className='bar'
+          grow
+          hAlignContent='left'
```
FrancescoCioria: `hAlignContent='left'` is the default -> please remove

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+  className: t.maybe(t.String),
+  style: t.maybe(t.Object)
+}), ({ min, max, steps }) => {
+  const minStartValue = Math.min(...steps.map(r => r.startValue));
+  const maxStartValue = Math.max(...steps.map(r => r.endValue));
+
+  return min === minStartValue && max === maxStartValue;
+}, 'Props');
+
+const computeResult = (current, min, max) => {
```
FrancescoCioria: you mean `computePercentage` ?

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+const Range = t.refinement(t.struct({
+  startValue: t.Number,
+  endValue: t.Number,
+  fillingColor: t.maybe(t.String),
+  labelColor: t.maybe(t.String)
+}), r => r.startValue < r.endValue, 'Range');
+
+const Ranges = t.refinement(t.list(Range), (rangeList) => {
+
+  const stepsWithout = (steps, stepToRemove) => {
```
FrancescoCioria: what does it mean `stepsWithout`?

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+    Math.max(step1.startValue, step2.startValue) < Math.min(step1.endValue, step2.endValue)
+  );
+
+  const noOverlappingRanges = (steps) => {
+    return every((steps), step1 => {
+      return every(stepsWithout(steps, step1), step2 => {
+        return !isOverlapped(step1, step2);
+      });
+    });
+  };
```
FrancescoCioria: this function is very ugly :stuck_out_tongue_closed_eyes: (three nested `return` ?)

src/meter/Meter.js
```diff
@@ -0,0 +1,161 @@
...
+      !(step.startValue === stepToRemove.startValue && step.endValue === stepToRemove.endValue)
+    ));
+  };
+
+  const isOverlapped = (step1, step2) => (
+    Math.max(step1.startValue, step2.startValue) < Math.min(step1.endValue, step2.endValue)
+  );
+
+  const noOverlappingRanges = (steps) => {
+    return every((steps), step1 => {
```
FrancescoCioria: why is `steps` between parenthesis?

src/meter/Meter.js
nerder: i use this as default if the user dosen't pass any of the relative props  and as a fall back when the user insert some kind of hole into the steps declaration (it's still possibile atm)

src/meter/Meter.js
nerder: Agree with that, i'll change in the next commit.

src/meter/Meter.js
nerder: Agree with that, i'll change in the next commit.

src/meter/Meter.js
nerder: Yes definitively it is. That props validation is actually a bit complex, and there no easy way of doing that. To check if a range is overlapped i have to compare each range with every other range except himself (all of this because rangeList is not sorted). If you have any refactoring proposal are  appreciated.

src/meter/Meter.js
nerder: my bad, fix in the next commit.

src/meter/Meter.js
FrancescoCioria: I know, what I'm saying is it makes no different having the HTML default `transparent` instead of yours arbitrary `#CCC` In both case it's a defect as it's very unlikely the user wants `#CCC`. That's why I suggest you replace this default with a warning

src/meter/Meter.js
```diff
@@ -0,0 +1,171 @@
...
+ */
+@skinnable()
+@props(Props)
+export default class Meter extends React.Component{
+
+  static defaultProps = {
+    min: 0,
+    max: 100,
+    defaultFillingColor: '#ccc',
+    defaultLabelColor: 'inherit',
```
FrancescoCioria: `inherit` is the HTML default, you can remove this (you don't have to give every prop a default value :))

src/meter/Meter.js
```diff
@@ -0,0 +1,171 @@
...
+
+const Range = t.refinement(t.struct({
+  startValue: t.Number,
+  endValue: t.Number,
+  fillingColor: t.maybe(t.String),
+  labelColor: t.maybe(t.String)
+}), r => r.startValue < r.endValue, 'Range');
+
+const Ranges = t.refinement(t.list(Range), (rangeList) => {
+  const rangeListNoColor = rangeList.map(r => ( { startValue: r.startValue, endValue: r.endValue }) );
```
giogonzo: useless

FrancescoCioria:
:tada: :balloon:

nerder:
:+1: :cake: :boom:

PR #269, opened 2016-02-13
#268: [FlexView] remove confusing `auto` prop (closes #268)

### LINKED ISSUE #268: [FlexView] remove confusing `auto` prop

FrancescoCioria:
Issue #268

### DIFF
--- a/src/flex/FlexView.js
+++ b/src/flex/FlexView.js
@@ -3,7 +3,6 @@ import cx from 'classnames';
 import pick from 'lodash/object/pick';
 import omit from 'lodash/object/omit';
 import { props, t } from '../utils';
-import { warn } from '../utils/log';
 
 
 const PropTypes = {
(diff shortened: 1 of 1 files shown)


src/flex/FlexView.js
```diff
@@ -76,6 +73,17 @@ const PropTypes = {
...
 export default class FlexView extends React.Component {
 
+  componentDidMount() {
+    this.logWarnings();
+  }
+
+  logWarnings = () => {
+    const { auto } = this.props;
+    if (!t.Nil.is(auto)) {
+      warn('"auto" is DEPRECATED. Use combination of "grow", "shrink" and "basis" instead');
```
giogonzo: we are not deprecating `auto` with this change, we are completely removing it :P

So this should be changed in something more strong, but probably could just be listed as breaking change in our CHANGELOG

FrancescoCioria: soo `"auto" is no longer supported. ...` ?

giogonzo: what I'm saying is it's not common to have warnings for features of old versions that are not there anymore (would be impossible to maintain such logs evolving a library..)

IMHO this should be explicit enough if we bump the semver and list the breaking changes somewhere, but also ccing @gabro here

gabro: semver + breaking change changelog :+1: 

provided someone is will to fix this in all our projects at once :P

FrancescoCioria: > provided someone is will to fix this in all our projects at once :P

volunteers? no one? ok... I'll do it... (you were referring to me anyway, weren't you? :( )
jokes apart, these changes should break only a very small portion of used `FlexView`s -> we should be able to fix them without too much effort

giogonzo: not necessarily at once, projects should fix their flexview version to what's more appropriate (i.e. sticking to the b-r-c one, not updated, for a while if that's better) ;)

PR #271, opened 2016-02-13
#270: [FlexView] drop support for `flexBasis` (closes #270)

FrancescoCioria:
Issue #270

based on #269

PR #273, opened 2016-02-13
#272: [FlexView] basis should always default to 'auto' even if grow={true} (closes #272)

### LINKED ISSUE #272: [FlexView] basis should always default to 'auto' even if grow={true}
Recent events showed how browsers are inconsistent about the computation of a flex element with a percentage `flex-basis`. (#132 #250)

## requirements
computed size should should be easy to predict and consistent in all major browsers.

## specs
`flex-basis` should always default to `'auto'` unless the user explicitly sets it a different value.

FrancescoCioria:
Issue #272

based on #271

### DIFF
--- a/examples/components/flex/Column.example
+++ b/examples/components/flex/Column.example
@@ -1,25 +1,15 @@
-const Example = React.createClass({
+class Example extends React.Component {
 
-  propTypes: {},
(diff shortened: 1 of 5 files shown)


giogonzo:
:+1:

PR #282, opened 2016-02-16
#281: Lint errors fix (closes #281)

### LINKED ISSUE #281: Lint errors fix
## requirements
Clean the code base

## specs

- run `npm run lint` it will fix a lot by itself
- fix the remaining issues

## Note

As you can see `travis-ci` will fail because of an `eslint` rule, specifically `prefer-es6-class`. It will happen because we still have some `React.createClass` instead of ES6 classes, as discussed offline with @giogonzo i can leave it like that atm. So the TBD points here are: we have to 'relax' ESlint (using rules ignoring comments for instance) or just remove that rule?

cc @gabro @FrancescoCioria

nerder:
Issue #281

## Test Plan

- run `npm run lint` and you will see no errors

NB: (there are still some errors but are not fixable, need more effort `React.createClass` into ES6 classes)

### DIFF
--- a/src/text-overflow/TextOverflow.js
+++ b/src/text-overflow/TextOverflow.js
@@ -53,7 +53,7 @@ export default class TextOverflow extends React.Component {
     const state = _state || this.state;
     if (state.isOverflowing === false) {
       const text = React.findDOMNode(this.refs.text);
-      if(text.offsetWidth < text.scrollWidth){
+      if (text.offsetWidth < text.scrollWidth){
         this.setState({ isOverflowing: true }, this.logWarnings);
       } else {
         this.logWarnings();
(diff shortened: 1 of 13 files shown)


src/text-overflow/TextOverflow.js
```diff
@@ -53,7 +53,7 @@ export default class TextOverflow extends React.Component {
     const state = _state || this.state;
     if (state.isOverflowing === false) {
       const text = React.findDOMNode(this.refs.text);
-      if(text.offsetWidth < text.scrollWidth){
+      if (text.offsetWidth < text.scrollWidth){
```
FrancescoCioria: missing space after before the `{`

gabro:
test

src/text-overflow/TextOverflow.js
nerder: no evidence of this in the current lint config

gabro:
Please rebase this on master after #284 is merged.

src/text-overflow/TextOverflow.js
gabro: you had this coming: https://github.com/buildo/eslint-config/pull/44

FrancescoCioria:
:+1:

PR #283, opened 2016-02-16
#280: [Button] no check before invoke setState() for still mounted component (closes #280)

### LINKED ISSUE #280: [Button] no check before invoke setState() for still mounted component
## description
Sometimes buttons, e.g.: Save for comment in sample details throw a warning when clicked.
Check against still mounted component with isMounted() is an antipattern, so a proper solution is to make the promise cancelable, although the warning seems caused by doResetInternalState() call.

## how to reproduce
- Have a button which `onClick` cause a re-render

## specs
- check component is still **mounted** when `setState` is called after the `onClick` promise resolution.

## misc
[optional: other useful info]

ascariandrea:
Issue #280

As pointed out [here](https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html) I made the `onClick` returning promise cancelable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "build": "rm -rf lib && mkdir lib && node-sass src --importer sass-importer.js --include-path node_modules -o lib && babel --loose --stage 0 --out-dir lib src",
     "lint": "eslint src",
     "lint-fix": "eslint src --fix",
-    "preversion": "npm run lint && npm run test && npm run build-examples",
+    "preversion": "npm run lint && npm run test",
     "prepublish": "npm run build",
     "build-kitchen-sink": "npm run clean && webpack --config kitchen-sink/webpack.config.build.js --progress",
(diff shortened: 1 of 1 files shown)


giogonzo:
as pointed out here https://github.omnilab.our.buildo.io/buildo/labonline/issues/1159#issuecomment-4843 , no need to introduce the concept of "cancelable promise" just for this button..

I'm more for: track the mount status in a  `_isMounted` or similar property on the component instance. Should be a three lines addition..

(also: travis complains!)

FrancescoCioria:
> (also: travis complains!)

`Button` has tests. @ascariandrea you should probably update them

ascariandrea:
@FrancescoCioria yep I fixed them!
@giogonzo I think as long our problem is related to **promises** we should stick to this `makeCancelable` implementation.

giogonzo:
@ascariandrea my opinion is: this problem is not related to promises in any way, but instead to any async thing we might be doing in a component. Fixing/changing the fact that promises are non-cancelable looks out of scope to me, and also this fixes it for this component only..

quick poll @buildo/frontend , opinions?

PR #284, opened 2016-02-16
Run lint before tests

gabro:
So that Travis catches linting issues

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+before_install:
+  - npm install -g npm@3
 node_js:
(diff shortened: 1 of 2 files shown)


FrancescoCioria:
:+1:

FrancescoCioria:
@gabro this way we're testing with last version of `npm3`, wouldn't it be better to force a version (like `npm@3.7.1`?

I'm saying this because for examples `npm@3.7.0` introduced major bugs that were resolved a week later with `3.7.1`

gabro:
Sure thing. Is 3.7.1 good?

PR #288, opened 2016-02-16
#287: [Panel] Panel Header should have overflow: hidden as default (closes #287)

### LINKED ISSUE #287: [Panel] Panel Header should have overflow: hidden as default
## description
When Panel Header content has a major width than the Panel Header itself it overflows.

## how to reproduce
- Put some content inside the Panel Header that has a min-width and stretch the panel to a lower width.

## specs
- add `overflow: hidden` to `.panel-header` class

ascariandrea:
Issue #287

### DIFF
--- a/src/Panel/panel.scss
+++ b/src/Panel/panel.scss
@@ -2,31 +2,31 @@
 
 @keyframes loading {
   0% {
-    left: 0;
     right: 100%;
+    left: 0;
   }
 
   50% {
-    left: 0;
     right: 0;
+    left: 0;
   }
 
   100% {
-    left: 100%;
     right: 0;
+    left: 100%;
   }
 }
 
 .panel {
   position: relative;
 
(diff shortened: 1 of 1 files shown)


src/Panel/panel.scss
```diff
@@ -35,42 +35,43 @@
...
 
     .expand-icon {
       font-size: 20px;
     }
 
     .panel-header-title {
-      font-weight: $semibold;
       font-size: 16px;
+      font-weight: $semibold;
+      padding-left: 20px;
```
ascariandrea: Move from `.panel-header` to `.panel-header-title` cause as long as `.panel` hasn't a `min-width` specified when it is shrank an area of 20px (`padding-left: 20px`) remains visible.

FrancescoCioria: :+1:

FrancescoCioria:
:100:

PR #290, opened 2016-02-16
#289: Reset hover on table rows hover (closes #289)

### LINKED ISSUE #289: Reset hover on table rows hover
To solve https://github.omnilab.our.buildo.io/buildo/labonline/issues/1616

francescogior:
Issue #289

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -37,10 +37,10 @@ $hover-scrollbar-color: white !default;
   }
 
   // bg color for odd/even rows (applied to cells for reliability)
-  .public_fixedDataTableRow_odd .fixedDataTableCellLayout_main {
+  .public_fixedDataTableRow_odd .public_fixedDataTableCell_main {
     background: $odd-row-background;
   }
-  .public_fixedDataTableRow_even .fixedDataTableCellLayout_main {
+  .public_fixedDataTableRow_even .public_fixedDataTableCell_main {
     background: $even-row-background;
   }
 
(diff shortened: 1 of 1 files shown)


src/table/table.scss
```diff
@@ -57,14 +57,14 @@ $hover-scrollbar-color: white !default;
 
   // selection/hover
   &.selectable {
-    .public_fixedDataTableCell_cellContent {
+    .public_fixedDataTableRow_main {
```
FrancescoCioria: If there is a row main class couldn't we use it below instead of row_odd and row_even?

src/table/table.scss
```diff
@@ -37,10 +37,10 @@ $hover-scrollbar-color: white !default;
   }
 
   // bg color for odd/even rows (applied to cells for reliability)
-  .public_fixedDataTableRow_odd .fixedDataTableCellLayout_main {
+  .public_fixedDataTableRow_odd .public_fixedDataTableCell_main {
```
FrancescoCioria: Is it normal the The public one is missing the suffix  `Layout`?

src/table/table.scss
francescogior: It's because this rule must override the default background, that is applied to row_odd and row_even

src/table/table.scss
francescogior: It's just another className, they both exist!

FrancescoCioria:
:+1

PR #293, opened 2016-02-17
#280: [Button] no check before invoke setState() for still mounted component (closes #280)

### LINKED ISSUE #280: [Button] no check before invoke setState() for still mounted component

ascariandrea:
Issue #280

After discussion in #283 I ended up with @giogonzo solution.
The tests are almost useless now, so maybe we need to rethink about them (in a separate pr).

I also defined global `it` and `describe` in the `.eslintrc` for tests lint.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,7 @@
 {
-  "extends": "buildo"
+  "extends": "buildo",
+  "globals": {
+    "describe": true,
+    "it": true,
+  }
 }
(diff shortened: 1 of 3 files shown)


giogonzo:
looks good :+1:  Tests are ugly / broken by design (not your fault)

PR #295, opened 2016-02-17
Automate showroom sync and deploy

FrancescoCioria:


FrancescoCioria:
automerging as related to examples

PR #297, opened 2016-02-17
#296: [Panel] Panel should have shrink=false for all its children (closes #296)

### LINKED ISSUE #296: [Panel] Panel should have shrink=false for all its children
The last change about `overflow: hidden;` for `panel-header` broke the PanelMenu visibility in open state.
The alternative solution could be set `shrink={false}` for `PanelHeader` and its inside element (`.panel-header-icon`, `.panel-header-title`, `.panel-header-content`) to make it take a `min-width` equals to its inside width.

@buildo/frontend thoughts?

ascariandrea:
Issue #296

### DIFF
--- a/src/Panel/PanelHeader.js
+++ b/src/Panel/PanelHeader.js
@@ -58,7 +58,7 @@ export default class PanelHeader extends React.Component {
           hAlignContent='center'
           style={{ cursor: 'pointer' }}
           className='panel-header-icon'
-          flexBasis={30}
+          basis={30}
           onClick={collapse.onToggleExpanded}
         >
           <Icon
@@ -73,7 +73,7 @@ export default class PanelHeader extends React.Component {
   templateTitle = ({ renderTitle, title }) => {
     return (
       renderTitle ?
-        <FlexView auto vAlignContent='center' className='panel-header-title'>
+        <FlexView auto vAlignContent='center' className='panel-header-title' shrink={false}>
           {title}
         </FlexView>
         : null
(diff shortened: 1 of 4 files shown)


src/Panel/panel.scss
```diff
@@ -60,12 +60,12 @@
   }
 
   > .panel-header {
-    overflow: hidden;
     color: $white;
     background-color: $charcoal;
 
     .expand-icon {
       font-size: 20px;
+      padding: 0 10px;
```
FrancescoCioria: Why? It already has `basis={30}` which is more than the width of the icon

src/Panel/PanelHeader.js
```diff
@@ -58,7 +58,8 @@ export default class PanelHeader extends React.Component {
           hAlignContent='center'
           style={{ cursor: 'pointer' }}
           className='panel-header-icon'
-          flexBasis={30}
+          basis={30}
+          shrink={false}
```
FrancescoCioria: This can be removed as shrink is already false when you set a basis

ascariandrea: :+1:

src/Panel/panel.scss
ascariandrea: I used padding before change `flexBasis` to `basis` and I forgot to remove it :see_no_evil:

FrancescoCioria:
👌

ascariandrea:
To be rebased after #302 merge.

FrancescoCioria:
merged #302, please rebase

ascariandrea:
Done :+1:

PR #304, opened 2016-02-18
#303: fix components using `FlexView` + `auto` (closes #303)

### LINKED ISSUE #303: fix components using `FlexView` + `auto`
## requirements
after #268 we should fix every components using `<FlexView auto ... />`

## specs
case by case fix using only `basis`, `grow` and `shrink`

FrancescoCioria:
Issue #303

### DIFF
--- a/src/Panel/Panel.js
+++ b/src/Panel/Panel.js
@@ -150,7 +150,7 @@ export default class Panel extends React.Component {
 
   templateExpandedContent = ({ children, loading }) => {
     return (
-      <FlexView className='panel-content' column grow auto style={{ position: 'relative' }}>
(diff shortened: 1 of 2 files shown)


ascariandrea:
I'm quite sure your fixes are correct, but I didn't have the knowledge of what `auto` achieved.
So :+1:, but next time please ade some specs to the related issue :)

FrancescoCioria:
> but next time please ade some specs to the related issue :)

the fact is, [those](https://github.com/buildo/react-components/issues/303) are the only possible specs as `auto` changed the defaults so it really depended on how and where it was used :)

PR #307, opened 2016-02-18
#305: [FlexView] basis should append 'px' also to string numbers (closes #305)

### LINKED ISSUE #305: [FlexView] basis should append 'px' also to string numbers
## requirements
we want `basis` to work exactly like `width` and `height`.
As `width='5'` is valid (it becomes `5px`) then also `basis='5'` should be valid and compute the same value (`5px`).

## specs
I can't see any clean way to do this as we **don't want** this to be valid too: `basis='5asdasd'`.
This means we can't simply use `parseInt`.
Still we could do this:
```js
const shouldAppendPx = t.Number.is(basis) || String(parseInt(basis, 10)) === basis;
...
```

FrancescoCioria:
Issue #305 

**test plan**
- `<FlexVIew basis="200" />` -> `flex: 0 0 200px`
- `<FlexVIew basis="200a" />` -> `flex: 0 0 200a`

### DIFF
--- a/src/flex/FlexView.js
+++ b/src/flex/FlexView.js
@@ -99,7 +99,7 @@ export default class FlexView extends React.Component {
   getBasis = () => {
     const { basis } = this.props;
     if (basis) {
-      const suffix = t.Number.is(basis) ? 'px' : '';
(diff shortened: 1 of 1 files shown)


francescogior:
:us:

PR #310, opened 2016-02-18
#277: TimePicker (closes #277)

### LINKED ISSUE #277: TimePicker
## Requirements
- Name: `TimePicker`
- There should be an input form
  - it can be manually filled
    - input will be validated
    - should support 1min interval
  - can be filled selecting from a dropdown list
    - each item of the list should have 30min interval
- The value can support both 24h and 12h time formats
- There should be a placeholder on empty values
- Values can be typed in in multiple ways

## Specs
- `TimePicker` is a self closing JSX tag, doesn’t accept any child
- It will wrap a  `Dropdown` from `b-r-c`
- `value` is an Object composed by `hours` and `minutes`
- When the user start typing the `Dropdown` will be filtered
- The dropdown list will include only times between [min..max]
- The interval (aka step value) is fixed to 30 min, because smaller intervals would lead to a poor UX
- Seconds are not handled
- Choose one of the item from the list will override user typed one
- The manually typed input have to be validated (if is not part of the option list already or it can be)
  - This is a validation for 12h
    - `^([1-9]|1[0-2])[:,\.\|\/\\]?([0-5]\d)?\s*?(AM|PM)$`
  - This is a validation for 24h
    - `^([01]\d|2[0-3])[:,\.\|\/\\]?([0-5]\d)?$`
- If the input validation failed, the input will become red ( or something)
- the placeholder will be a string that not have to be validated

## API [Draft]
```javascript
<TimePicker 
  value={Object}
  min={Object}
  max={Object}
  format={Enum}
  placeholder={String}
/>
```
## Defaults
**value**: `undefined`
**min** : `undefined`
**max** : `undefined`
**format** : `24h`
**placeholder** : ``

nerder:
Issue #277

### DIFF
--- a/examples/components/time-picker/TimePicker12.example
+++ b/examples/components/time-picker/TimePicker12.example
@@ -9,7 +9,7 @@ class Example extends React.Component {
     return (
       <div>
         <h1>12 Hour Time Picker</h1>
-        <TimePicker value={this.state.value} onChange={(value) => this.setState({ value })} interval={30} timeFormat="12h" />
+        <TimePicker value={this.state.value} onChange={(value) => this.setState({ value })} timeFormat="12h" />
       </div>
     );
   }
--- a/examples/components/time-picker/TimePicker12.example
+++ b/examples/components/time-picker/TimePicker12.example
@@ -1,0 +1,11 @@
+class Example extends React.Component {
+
+  render = () => {
+    return (
+      <div>
+        <h1>12 Hour Time Picker</h1>
+        <TimePicker onChange={(x) => log(x)} interval={30} timeFormat="12h" />
+      </div>
+    );
+  }
+}
--- a/examples/components/time-picker/TimePicker12.example
+++ b/examples/components/time-picker/TimePicker12.example
@@ -2,7 +2,7 @@ class Example extends React.Component {
 
   constructor() {
     super();
-  	this.state = {};
+    this.state = {};
   }
 
   render = () => {
--- a/examples/components/time-picker/TimePicker12.example
+++ b/examples/components/time-picker/TimePicker12.example
@@ -1,10 +1,15 @@
 class Example extends React.Component {
 
+  constructor() {
+    super();
+  	this.state = {};
+  }
+
   render = () => {
     return (
       <div>
         <h1>12 Hour Time Picker</h1>
-        <TimePicker onChange={(x) => log(x)} interval={30} timeFormat="12h" />
+        <TimePicker value={this.state.value} onChange={(value) => this.setState({ value })} interval={30} timeFormat="12h" />
       </div>
     );
   }
--- a/examples/components/time-picker/TimePicker24.example
+++ b/examples/components/time-picker/TimePicker24.example
@@ -1,10 +1,15 @@
 class Example extends React.Component {
 
+  constructor() {
+    super();
+  	this.state = {};
+  }
+
   render = () => {
     return (
       <div>
         <h1>24 Hour Time Picker</h1>
-        <TimePicker onChange={(x) => log(x)} interval={30} timeFormat="24h" />
+        <TimePicker value={this.state.value} onChange={(value) => this.setState({ value })} interval={30} timeFormat="24h" />
       </div>
     );
   }
--- a/examples/components/time-picker/TimePicker24.example
+++ b/examples/components/time-picker/TimePicker24.example
@@ -6,10 +6,12 @@ class Example extends React.Component {
   }
 
   render = () => {
+    const minTime = { hours : 8, minutes: 30 };
+    const maxTime = { hours : 18, minutes: 30 };
     return (
       <div>
         <h1>24 Hour Time Picker</h1>
-        <TimePicker value={this.state.value} onChange={(value) => this.setState({ value })} interval={30} timeFormat="24h" />
+        <TimePicker value={this.state.value} minTime={minTime} maxTime={maxTime} onChange={(value) => this.setState({ value })} interval={30} timeFormat="24h" />
       </div>
     );
   }
--- a/examples/components/time-picker/TimePicker24.example
+++ b/examples/components/time-picker/TimePicker24.example
@@ -11,7 +11,7 @@ class Example extends React.Component {
     return (
       <div>
         <h1>24 Hour Time Picker</h1>
-        <TimePicker value={this.state.value} minTime={minTime} maxTime={maxTime} onChange={(value) => this.setState({ value })} interval={30} timeFormat="24h" />
+        <TimePicker value={this.state.value} minTime={minTime} maxTime={maxTime} onChange={(value) => this.setState({ value })} timeFormat="24h" />
       </div>
     );
   }
--- a/examples/components/time-picker/TimePicker24.example
+++ b/examples/components/time-picker/TimePicker24.example
@@ -1,0 +1,11 @@
+class Example extends React.Component {
+
+  render = () => {
+    return (
+      <div>
+        <h1>24 Hour Time Picker</h1>
+        <TimePicker onChange={(x) => log(x)} interval={30} timeFormat="24h" />
+      </div>
+    );
+  }
+}
--- a/examples/components/time-picker/TimePicker24.example
+++ b/examples/components/time-picker/TimePicker24.example
@@ -2,7 +2,7 @@ class Example extends React.Component {
 
   constructor() {
     super();
-  	this.state = {};
+    this.state = {};
   }
 
   render = () => {
--- a/examples/components/time-picker/default.example
+++ b/examples/components/time-picker/default.example
@@ -1,11 +1,13 @@
 class Example extends React.Component {
 
+  logChange = (val) => {
+    console.log('Selected val =>', val);
+  }
+
   render = () => {
     const Time = { hours: 10, minutes: 59 };
-    const minTime = { hours: 6, minutes: 20 };
-    const maxTime = { hours: 15, minutes: 20 };
     return (
-      <TimePicker value={Time} minTime={minTime} maxTime={maxTime} timeFormat="24h" />
+      <TimePicker value={Time} interval={30} timeFormat="12h" onChange={this.logChange} />
     );
   }
 }
--- a/examples/components/time-picker/default.example
+++ b/examples/components/time-picker/default.example
@@ -1,13 +1,11 @@
 class Example extends React.Component {
 
-  logChange = (val) => {
-    console.log('Selected val =>', val);
-  }
-
   render = () => {
     const Time = { hours: 10, minutes: 59 };
+    const Min = { hours: 8, minutes: 25 };
+    const Max = { hours: 12, minutes: 32 };
     return (
-      <TimePicker value={Time} interval={30} timeFormat="12h" onChange={this.logChange} />
+      <TimePicker value={Time} onChange={(x) => log(x)}maxTime={Max} minTime={Min} interval={30} timeFormat="24h"/>
     );
   }
 }
--- a/examples/components/time-picker/default.example
+++ b/examples/components/time-picker/default.example
@@ -1,5 +1,11 @@
 class Example extends React.Component {
-  render = () => (
-    <TimePicker />
-  );
+
+  render = () => {
+    const Time = { hours: 10, minutes: 59 };
+    const minTime = { hours: 6, minutes: 20 };
+    const maxTime = { hours: 15, minutes: 20 };
+    return (
+      <TimePicker value={Time} minTime={minTime} maxTime={maxTime} timeFormat="24h" />
+    );
+  }
 }
--- a/examples/components/time-picker/default.example
+++ b/examples/components/time-picker/default.example
@@ -1,11 +1,8 @@
 class Example extends React.Component {
 
   render = () => {
-    const Time = { hours: 10, minutes: 59 };
-    const Min = { hours: 8, minutes: 25 };
-    const Max = { hours: 12, minutes: 32 };
     return (
-      <TimePicker value={Time} onChange={(x) => log(x)}maxTime={Max} minTime={Min} interval={30} timeFormat="24h"/>
+      <TimePicker onChange={(x) => log(x)} interval={30} timeFormat="24h" />
     );
   }
 }
--- a/examples/components/time-picker/default.example
+++ b/examples/components/time-picker/default.example
@@ -1,0 +1,5 @@
+class Example extends React.Component {
+  render = () => (
+    <TimePicker />
+  );
+}
--- a/examples/components/time-picker/default.example
+++ b/examples/components/time-picker/default.example
@@ -1,8 +1,0 @@
-class Example extends React.Component {
-
-  render = () => {
-    return (
-      <TimePicker onChange={(x) => log(x)} interval={30} timeFormat="24h" />
-    );
-  }
-}
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -10,34 +10,32 @@ const H24 = '24h';
 const H12 = '12h';
 const symbolRegex = /[\s,\.\|\/\\]/;
 const numberRegex = /^\d+$/;
-
+const separator = ':';
 
 const pad = (num) => num <= 9 ? `0${num}` : num;
-
 const timeFormatter24 = (hour, minute) => `${pad(hour)}:${pad(minute)}`;
 const timeFormatter12 = (hour, minute) => {
   if (hour === 0) {
-    return `12:${pad(minute)} am`;
+    return `12${separator}${pad(minute)} am`;
   } else if (hour === 12) {
-    return `12:${pad(minute)} pm`;
+    return `12${separator}${pad(minute)} pm`;
   } else if (hour >= 13) {
-    return `${pad(hour % 12)}:${pad(minute)} pm`;
+    return `${pad(hour % 12)}${separator}${pad(minute)} pm`;
   } else {
-    return `${pad(hour)}:${pad(minute)} am`;
+    return `${pad(hour)}${separator}${pad(minute)} am`;
   }
 };
 
 const insertColon = str => {
   if (numberRegex.test(str) && (str.length >= 3 && str.length <= 4)) {
     const position = str.length - 2;
-    const strWithColon = `${str.substr(0, position)}:${str.substr(position)}`;
+    const strWithColon = `${str.substr(0, position)}${separator}${str.substr(position)}`;
     return strWithColon;
   }
   return str;
 };
 const cleanString = str => str.replace(symbolRegex, ':');
 const cleanFilter = filterStr => insertColon(cleanString(filterStr));
-
 const filterOptions = (options, filterStr) => (
   options.filter( option => (
       option.value.substr(0, filterStr.length) === cleanFilter(filterStr) ||
@@ -51,14 +49,11 @@ const Integer = t.refinement(t.Number, n => n % 1 === 0, 'Integer');
 const Hour = t.refinement(Integer, int => int >= 0 && int <= 23, 'Hour');
 const Hour12 = t.refinement(Hour, int => int >= 1 && int <= 12, 'Hour12');
 const Minute = t.refinement(Integer, int => int >= 0 && int <= 59, 'Minute');
-
 const TimeFormat = t.enums.of([H12, H24], 'TimeFormat');
-
 const Time = t.struct({
   hours: Hour,
   minutes: Minute
 }, 'Time');
-
 const Time12 = t.struct({
   hours: Hour12,
   minutes: Minute
@@ -66,12 +61,12 @@ const Time12 = t.struct({
 
 const Props = t.refinement(t.struct({
   value: Time,
+  onChange: t.Function,
   minTime: t.maybe(Time),
   maxTime: t.maybe(Time),
   interval: t.maybe(Integer),
   placeholder: t.maybe(t.String),
   timeFormat: t.maybe(TimeFormat),
-  onChange: t.maybe(t.Function),
   id: t.maybe(t.String),
   className: t.maybe(t.String),
   style: t.maybe(t.Object)
@@ -84,8 +79,8 @@ const Props = t.refinement(t.struct({
 export default class TimePicker extends React.Component {
 
   static defaultProps = {
-    placeholder: '--:--',
-    timeFormat: '24h',
+    placeholder: `--${separator}--`,
+    timeFormat: H24,
     interval: 30,
     onChange: () => {}
   }
@@ -97,11 +92,18 @@ export default class TimePicker extends React.Component {
     const hours = range(minHour, maxHour);
     const minutes = range(0, 60, interval);
     const options = flatten(hours.map( hour => minutes.map( minute => (
-      { value: `${hour}:${minute}`, label: timeFormatter(hour, minute) }
+      { value: `${hour}${separator}${minute}`, label: timeFormatter(hour, minute) }
     )
     )));
     return options;
   };
+
+  _onChange = (value) => {
+    if (value) {
+      const timeArray = value.split(separator);
+      this.props.onChange({ hours: parseInt(timeArray[0]), minutes: parseInt(timeArray[1]) });
+    }
+  }
 
   getLocals() {
     const {
@@ -114,7 +116,7 @@ export default class TimePicker extends React.Component {
 
     return {
       placeholder,
-      onChange,
+      onChange: this._onChange,
       filterOptions,
       className: cx('time-picker', className),
       options: timeFormat === H24 ? this.createOptionsList({ timeFormatter: timeFormatter24, ...props }) : this.createOptionsList({ timeFormatter: timeFormatter12, ...props })
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -26,7 +26,7 @@ const timeFormatter12 = (hour, minute) => {
   }
 };
 
-const compareTime = (minTime, maxTime) => (
+const lteTime = (minTime, maxTime) => (
   maxTime.hours > minTime.hours || (maxTime.hours === minTime.hours && maxTime.minutes > minTime.minutes)
 );
 
@@ -70,7 +70,7 @@ const Props = t.refinement(t.struct({
   className: t.maybe(t.String),
   style: t.maybe(t.Object)
 }), ({ value, minTime, maxTime }) => (
-  compareTime(minTime, maxTime) && (!value || (compareTime(value, maxTime) && compareTime(minTime, value)))
+  lteTime(minTime, maxTime) && (!value || (lteTime(value, maxTime) && lteTime(minTime, value)))
 ), 'Props');
 
 @skinnable()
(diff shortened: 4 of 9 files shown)


src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,101 @@
...
+import cx from 'classnames';
+import Dropdown from '../dropdown/Dropdown';
+import FlexView from '../flex/FlexView';
+import range from 'lodash/utility/range';
+import flatten from 'lodash/array/flatten';
+
+const H24 = '24h';
+const H12 = '12h';
+
+const pad = (num) => ( num <= 9 ? `0${num}` : num );
```
giogonzo: no parens here

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,101 @@
...
+
+const timeFormatter24 = (hour, minute) => (`${pad(hour)}:${pad(minute)}`);
```
giogonzo: no parens here as well

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,101 @@
...
+    return `12:${pad(minute)} pm`;
+  } else if (hour >= 13) {
+    return `${pad(hour % 12)}:${pad(minute)} pm`;
+  } else {
+    return `${pad(hour)}:${pad(minute)} am`;
+  }
+};
+
+const Integer = t.refinement(t.Number, n => n % 1 === 0, 'Integer');
+const Hour = t.refinement(Integer, int => int >= 1 && int <= 23, 'Hour');
```
giogonzo: aren't you missing either `0` or `24`?

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,101 @@
...
+  maxTime: t.maybe(Time),
+  interval: t.maybe(Integer),
+  placeholder: t.maybe(t.String),
+  timeFormat: t.maybe(TimeFormat),
+  onChange: t.maybe(t.Function),
+  id: t.maybe(t.String),
+  className: t.maybe(t.String),
+  style: t.maybe(t.Object)
+}), ({ timeFormat, value, minTime, maxTime }) => (
+  timeFormat === H24 || (Time12(value) && (!minTime || Time12(minTime)) && (!maxTime || Time12(maxTime)))
```
giogonzo: still ugly, but you should be able to replace `(!minTime || Time12(minTime))` with `t.maybe(Time12)(minTime)` and `(!maxTime || Time12(maxTime))` with `t.maybe(Time12)(maxTime)`

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,101 @@
...
+  static defaultProps = {
+    placeholder: '--:--',
+    timeFormat: '24h',
+    interval: 30,
+    onChange: () => {}
+  }
+
+  createOptionsList = ({ timeFormatter, minTime, maxTime, interval }) => {
+
+    const hours = range(minTime.hours, maxTime.hours);
```
giogonzo: what if `minTime` or `maxTime` are not provided?

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,101 @@
...
+    const minutes = range(0, 60, interval);
```
giogonzo: what about `minTime.minutes` and `maxTime.minutes`?

src/time-picker/TimePicker.js
nerder: doesn't seem to work as expected.

src/time-picker/TimePicker.js
nerder: i didn't implement this yet because is not 100% clear how i can actually implement this. TBD

src/time-picker/TimePicker.js
giogonzo: right I see why now, forget it

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,139 @@
...
+  minTime: t.maybe(Time),
+  maxTime: t.maybe(Time),
+  interval: t.maybe(Integer),
+  placeholder: t.maybe(t.String),
+  timeFormat: t.maybe(TimeFormat),
+  id: t.maybe(t.String),
+  className: t.maybe(t.String),
+  style: t.maybe(t.Object)
+}), ({ value, minTime, maxTime }) => (
+  compareTime(minTime, maxTime) && (!value || (compareTime(value, maxTime) && compareTime(minTime, value)))
```
giogonzo: `compareTime` is not explicative as name for this function IMO
something like `lte` or `lowerThanOrEqual` would be better

nerder: :+1:

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,139 @@
...
+  createOptionsList = ({ timeFormatter, minTime, maxTime, interval }) => {
+
+    const minHour = minTime.hours;
+    const maxHour = maxTime.hours;
+    const minMinute = minTime.minutes;
+    const maxMinute = maxTime.minutes;
+
+    const hours = range(minHour, maxHour + 1);
+    const minutes = range(0, 60, interval);
+    const options = flatten(hours.map( hour => minutes.filter( minute => (
```
giogonzo: eslint won't complain, but no spaces before `hour` and `minute`

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,139 @@
...
+      (hour !== minHour || minute >= minMinute) &&
+      (hour !== maxHour || minute <= maxMinute)
+    )).map( minute => (
```
giogonzo: same here

also, I find it more readable written as

```js
.map(minute => ({
  value: ...,
  label: ...
}))
```

examples/components/time-picker/default.example
```diff
@@ -0,0 +1,8 @@
+class Example extends React.Component {
```
FrancescoCioria: this example should be improved to store in `this.state` the `value` returned by `onChange` and pass it to `TimePicker`

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+const H24 = '24h';
+const H12 = '12h';
+const symbolRegex = /[\s,\.\|\/\\]/;
+const numberRegex = /^\d+$/;
+const separator = ':';
+
+const pad = (num) => num <= 9 ? `0${num}` : num;
+const timeFormatter24 = (hour, minute) => `${pad(hour)}:${pad(minute)}`;
+const timeFormatter12 = (hour, minute) => {
+  if (hour === 0) {
```
FrancescoCioria: did you forget to change this? :stuck_out_tongue_closed_eyes:

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+  } else if (hour === 12) {
+    return `12${separator}${pad(minute)} pm`;
+  } else if (hour >= 13) {
+    return `${pad(hour % 12)}${separator}${pad(minute)} pm`;
+  } else {
+    return `${pad(hour)}${separator}${pad(minute)} am`;
+  }
+};
+
+const lteTime = (minTime, maxTime) => (
```
FrancescoCioria: `lteTime` ?

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+  maxTime.hours > minTime.hours || (maxTime.hours === minTime.hours && maxTime.minutes > minTime.minutes)
+);
+
+const insertColon = str => {
+  if (numberRegex.test(str) && (str.length >= 3 && str.length <= 4)) {
```
FrancescoCioria: maybe replace `(str.length >= 3 && str.length <= 4)` with `(str.length === 3 || str.length === 4)`? It would be easier to read IMO

src/time-picker/TimePicker.js
nerder: yep @giogonzo propose `lte` because `compareTime` doesn't looks very clear, at actually i'm agree with this, maybe without `Time`? Is defo a `lte` function, we should find out a better name

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+    const position = str.length - 2;
+    const strWithColon = `${str.substr(0, position)}${separator}${str.substr(position)}`;
+    return strWithColon;
+  }
+  return str;
+};
+const cleanString = str => str.replace(symbolRegex, ':');
```
FrancescoCioria: `cleanString`  doesn't communicate much. Shouldn't this be something like `useDefaultSeparator` or `cleanSeparator` or....?

src/time-picker/TimePicker.js
nerder: Ehm, actually yes :smile: i'll refactor this ASAP in the next commit.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+const cleanFilter = filterStr => insertColon(cleanString(filterStr));
+const filterOptions = (options, filterStr) => (
```
FrancescoCioria: `filterStr` is the user input right? maybe rename in `inputStr`? took me a while to understand it

src/time-picker/TimePicker.js
nerder: :+1: way better, hours can be a `float` than ok

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+  static defaultProps = {
+    placeholder: `--${separator}--`,
+    timeFormat: H24,
+    interval: 30,
+    minTime: { hours: 0, minutes: 0 },
+    maxTime: { hours: 23, minutes: 59 }
+  }
+
+  createOptionsList = ({ timeFormatter, minTime, maxTime, interval }) => {
+
```
FrancescoCioria: remove empty line :stuck_out_tongue_closed_eyes:

src/time-picker/TimePicker.js
nerder: +1 for `cleanSeparator`

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+    )).map( minute => ({
+      value: `${hour}${separator}${minute}`,
+      label: timeFormatter(hour, minute)
+    })
+    )));
+    return options;
+  };
+
+  _onChange = (value) => {
+    if (value) {
```
FrancescoCioria: if `value === ''` `this.props.onChange` is never called --> use can't delete the last character!

src/time-picker/TimePicker.js
nerder: +1 for `inputStr` is more clear

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+      className,
+      ...props
+    } = this.props;
+
+    const timeFormatter = timeFormat === H24 ? timeFormatter24 : timeFormatter12;
+
+    return {
+      ...props,
+      onChange: this._onChange,
+      filterOptions,
```
FrancescoCioria: why is `filterOptions` outside of the component `class`?

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+      className: cx('time-picker', className),
+      options: this.createOptionsList({ timeFormatter, ...props })
+    };
+  }
+
+  template({ id, className, style, value, placeholder, options, onChange, filterOptions }) {
+    return(
+      <FlexView {...{ id, className, style }} grow >
```
FrancescoCioria: please remove `grow`, the layout size it's its parent duty

FrancescoCioria: actually, please remove `FlexView`, you don't need it (pass `style, id, className` directly to `Dropdown`)

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  createOptionsList = ({ timeFormatter, minTime, maxTime, interval }) => {
+
+    const minHour = minTime.hours;
+    const maxHour = maxTime.hours;
+    const minMinute = minTime.minutes;
+    const maxMinute = maxTime.minutes;
+
+    const hours = range(minHour, maxHour + 1);
```
FrancescoCioria: here you use `hours` (plural) for an array while in `value.hours`, `minTime.hours` and `maxTime.hours` is used for a single value.

src/time-picker/TimePicker.js
FrancescoCioria: sorry, I didn't know `lte` standed for `less than equal`

src/time-picker/TimePicker.js
nerder: There's no reason for this, i was wondering what have to be inside and what can be outside instead, there is a general rules for this? My guess is that, all the thing that go inside of `getLocals()` have to stay inside, the utility function can stay outside. Am i wrong?

src/time-picker/TimePicker.js
nerder: way better to refactor `minHour` => `minHours` (same for max)

src/time-picker/TimePicker.js
FrancescoCioria: > My guess is that, all the thing that go inside of `getLocals()` have to stay inside

In general, yes. (so I would change it to **should** stay inside)

ex: also `timeFormatter` should be inside you component but, as you're using it as a `defaultProp` it must be statically defined --> it can't be inside your `class`

src/time-picker/TimePicker.js
nerder: actually this is intended in order to avoid to return Object bad formatted like: `Object {hours: NaN, minutes: NaN}`. Than i check if the value is not present i'll not call the onChange. Maybe i can implement an `else` statement that just call `this.props.onChange(value)` when the `value` is not well formatted.

FrancescoCioria: sorry, this is the `value` of `react-select` so it's called only when the user **clicks** on an option or if he clears it --> `value` can never be `''` but it can be `undefined`.

> Maybe I can implement an `else` statement that just call `this.props.onChange(value)` when the `value` is not well formatted.

yes, but when the `value` is `undefined` (which is falsy so it's ok a simple `else`)

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+      const position = str.length - 2;
+      const strWithColon = `${str.substr(0, position)}${separator}${str.substr(position)}`;
+      return strWithColon;
+    }
+  }
+  return str;
+};
+
+const cleanSeparator = str => str.replace(symbolRegex, separator);
+const cleanInput = inputStr => addZero(insertColon(cleanSeparator(inputStr)));
```
nerder: That function can be very unclear. What it actually do is to cleaning the input right before to be formatted, and in order to be manipulated well. It will cover many use cases, and each function involved in this will cover one of them.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+const formatValue = (timeFormatter, hour, minute) => ({
+  value: `${hour}${separator}${minute}`,
+  label: timeFormatter(hour, minute)
+});
+
+const lteTime = (minTime, maxTime) => (
+  maxTime.hours > minTime.hours || (maxTime.hours === minTime.hours && maxTime.minutes >= minTime.minutes)
+);
+
+const addZero = str => {
```
nerder: This function add a `0` at the end of every string that is already been cleaned and it is in the format `HH:m` in any other cases it will do nothing at all. It will be called just at the very end, and is useful to manage the decision we make to add a `0` when we are into an undefined situation like that. We can say that this is a "By Design behaviour"

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+  if (str.length === 4) {
+    return `${str}0`;
+  } else {
+    return str;
+  }
+};
+
+const insertColon = str => {
```
nerder: `insertColon` (where the name have to be refactored into `insertSeparator` IMHO) it's useful to handle another use case wanted by the requirements, so when the input have no separator. It will insert the separator (by default `:`) right in the middle of hours and minute. The code of this function is not very clear i guess, i'm starting implementing this from a code snippet i found. It can be refactored.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+    } else if (str.length === 4) {
+      const position = str.length - 2;
+      const strWithColon = `${str.substr(0, position)}${separator}${str.substr(position)}`;
+      return strWithColon;
+    }
+  }
+  return str;
+};
+
+const cleanSeparator = str => str.replace(symbolRegex, separator);
```
nerder: pretty easy function, it just will replace any white listed separator into the default/choosen one.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+    maxTime: { hours: 23, minutes: 59 }
+  }
+
+  /**
+    We use filterOptions because seems that is not possible
+    to change the options from outside of this function,
+    then we use this tricky hack and we workaround this creating a new set
+    of options as we wish and filter it if needed, or just returning a brand new array
+  **/
+  createOptionsList = (_, inputStr) => {
```
nerder: This is actually commented in the code base as well, because that function works on top of an hack we have to make, and this is because the version of react-select we are using will not support changing `options` in other ways. The main point of misunderstanding here is that it will return an array that actually can be 3 different things.

1) `undefined` case (the user just focus the input): pre generated list, with all the needed results computed using min,max,interval and everything we need.
2) The filtered list, just when the input is that format `H`, `HH`, `HH:` actually the options list will be regenerated completely and filtered.
3) In all the other valid use format like: `HH:m`,`HH:mm`, `HHm`, `HHmm` the list will be a completely new one, generated on the basis of timeFormat (12H,24H).

Refactor proposal:
- Naming `createOptionsList` dosent means nothing to me right now. Something better can be optionsManager or something like that. 
- It's wired that a single function return another blackbox function that returns arrays that you don't know how are they generated (one time filtered, one time generated, one time both)
- performance issues, is not SO expensive in terms of performance cause the arrays of options (when it need to be generated in the case 1) can have maximum a size of 48, but is just pointless to recreate it even when is not needed. That logic can be moved into a dedicated function.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+    const minutes = range(0, 60, interval);
+    const options = flatten(hours.map(hour => minutes.filter(minute => (
+      (hour !== minHours || minute >= minMinutes) &&
+      (hour !== maxHours || minute <= maxMinutes)
+    )).map( minute => formatValue(timeFormatter, hour, minute)
+    )));
+    return this.filterOrRecreateOptions(options, inputStr);
+  };
+
+  filterOrRecreateOptions = (options, inputStr) => {
```
nerder: This is the function described above, can be obviously splitted in to function, on for the filter on the other one for the custom generated optionsList.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+      return strWithColon;
+    }
+  }
+  return str;
+};
+
+const cleanSeparator = str => str.replace(symbolRegex, separator);
+const cleanInput = inputStr => addZero(insertColon(cleanSeparator(inputStr)));
+
+const getSpecificOptionList = (cleanedInput, timeFormat) => {
```
nerder: This is acutally the wirdest function here, because it will return the custom generated option list, that is different for 24h and 12h time formats. Of course 12h need to have an array of 2 options because we don't know since the very end (when he will click on one option) what he intends. This is actually a very custom behaviour for 12h. It works in a different way from what the formatter actually expect, because we don't have 2 different value (i.e. 11 & 23 => 11am & 11pm OR 0 & 12 => 12 am & 12pm ) but we have just the parsed value from the user input, not a generated one, and by design in can be always less or equal than 12, but as i said before we have to show the user both the choices. They looks similar but they are actually 2 completely different way of formatting. In order to use the already implemented we can use a simple solution, that is sum 12 to the input (that for sure is <= 12) this will return the exact opposite of the user input value, returning the results we need. But a very big issue here is that for managing this we have to expand the logic of the formatter for managing the special case of 24. This logic seems very complex, and actually it his, but i didnt have any other ideas that can smartly solve this problem without repeating code implementing 2 different formatter. Will be easier to understand reading the code.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+const H12 = '12h';
+const symbolRegex = /[\s,\.\|\/\\]/;
+const numberRegex = /^\d+$/;
+const time12Regex = /^([1-9]|1[0-2])[:,\.\|\/\\]?([0-5]\d)?\s*?(AM|PM)$/;
+const time24Regex = /^([01]?\d|2[0-3])[:,\.\|\/\\]?([0-5]?\d)?$/;
+const separator = ':';
+
+const pad = (num) => num <= 9 ? `0${num}` : num;
+const timeFormatter24 = (hour, minute) => `${pad(hour)}:${pad(minute)}`;
+const timeFormatter12 = (hour, minute) => {
```
nerder: This function is the very bad. Especially the first if statement, this is because there is more than one logic that have to be checked because it can have as input different things, it looks very specific but actually it work very generally. The first line manage the 0 case (in the optionList generation) and the 12+12 case for the custom generated list.
The second if can be removed even if i make a lot of tries and i didn't find out a way of remove it. The other 2 conditions are pretty obvious.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
...
+    } = this.props;
+
+    const timeFormatter = timeFormat === H24 ? timeFormatter24 : timeFormatter12;
+    return {
+      ...props,
+      className: cx('time-picker', className),
+      value: value ? formatValue(timeFormatter, value.hours, value.minutes) : '',
+      options,
+      onChange: this._onChange,
+      filterOptions: this.createOptionsList
```
nerder: filterOptions call something called `createOptionList` this is veeery unclear.

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,211 @@
+import React from 'react';
+import { props, t, skinnable } from '../utils';
+import cx from 'classnames';
+import Dropdown from '../dropdown/Dropdown';
+import range from 'lodash/utility/range';
+import flatten from 'lodash/array/flatten';
+
+/*
+  Options is defined here cause we need to not change the ref.
+  This happend because of a very wired behaviour of react-select v0.6.12
```
giogonzo: typos ^

however, a comment like `hardcoded empty options. We are generating options dynamically` would be more useful I think

src/time-picker/TimePicker.js
giogonzo: the first `if` condition simplifies to `hour === 0 || hour === 24`

the last `else` is useless (you're `% 12` anyway before)

```js
if (hour === 0 || hour === 24) { // WTF: why can we have both 0 and 24 in our internal representation? we should just choose one?
// ...
} else if (hour === 12) {
// ...
} else {
// ...
}
```

which makes me think: why can we have both `0` and `24` in our internal representation? we should just choose one?

nerder:
Now the codebase looks way more clear, as suggested from you i split the logic into a singular and modular part, this will hardly improve the testability of the component and the readability of the whole code base. I leaved a big comment cause i'm having an issue with the validation part mostly because i have to duplicate code, and it not seems fair to me, but i wasn't able to find out an intelligent fix for this. This looks like the last part that need to be refactored, plus i guess so destructuring will be useful as well. I'm waiting for you to review this refactor and i will start with unit testing asap.

/cc @giogonzo @FrancescoCioria 
:sleepy: :+1:

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,221 @@
...
+    const filteredTimeList = timesFilter(timeList, time, minTime, maxTime);
+    return filteredTimeList.map(time => formatter(time, timeFormat));
+  };
+
+  _onChange = (value) => {
+    if (value) {
+      const {
+        hours,
+        minutes
+      } = this.parser(value);
```
giogonzo: WAT?

nerder:
@FrancescoCioria Ok, this is my last commit, waiting for you last comments if you find out something that can be fixed :)

examples/components/time-picker/TimePicker12.example
```diff
@@ -0,0 +1,16 @@
+class Example extends React.Component {
+
+  constructor() {
+    super();
+  	this.state = {};
```
FrancescoCioria: wrong indent

examples/components/time-picker/TimePicker24.example
```diff
@@ -0,0 +1,18 @@
+class Example extends React.Component {
+
+  constructor() {
+    super();
+  	this.state = {};
```
FrancescoCioria: wrong indent

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,226 @@
...
+export const H24 = '24h';
+export const H12 = '12h';
+export const inputError = {};
+const symbolRegex = /[\s,\-\+\.\|\/\\]/;
+const numberRegex = /^\d+$/;
+const separator = ':';
+const interval = 30;
+
+const pad = (num) => num <= 9 ? `0${num}` : num;
+const hoursInH12 = hour => hour + 12 !== 24 ? hour + 12 : 0;
```
FrancescoCioria: this function does not return a range of H12 values (`[1, 12]`). maybe find a better name?

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,226 @@
...
+const TimeFormat = t.enums.of([H12, H24], 'TimeFormat');
+const Time = t.struct({
+  hours: Hour,
+  minutes: Minute
+}, 'Time');
+
+const isValidHoursInTimeFormat = (hours, timeFormat) => timeFormat === H24 ? Hour.is(hours) : Hour12.is(hours);
+
+const Props = t.refinement(t.struct({
+  onChange: t.Function,
```
FrancescoCioria: missing docs for `onChange`

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,226 @@
...
+    return { hours, minutes, originalInput: cleanedInput };
+  } else if (validHours) {
+    return { originalInput: cleanedInput };
+  } else {
+    return inputError;
+  }
+};
+
+export const createTimeList = ({ hours, minutes }, timeFormat) => {
+  if (!isValidHoursInTimeFormat(hours) || !Minute.is(minutes)) {
```
FrancescoCioria: `isValidHoursInTimeFormat` requires two arguments

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,226 @@
...
+    const hoursList = range(0, 24);
+    const minutesList = range(0, 60, interval);
+    return flatten(hoursList.map(hours => minutesList.map(minutes => ({ hours, minutes, timeFormat })
+    )));
```
FrancescoCioria: very weird indent

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,226 @@
...
+};
+
+export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
+  return lteTime(minTime, time) && lteTime(time, maxTime) && startsWith(time, originalInput);
+};
+
+// we are not "filtering" options (options array is always empty and discarded by this function)
+// our `filterOptions` actually generates options based on current input string and props.
+// We use this as a workaround because there's not any other easy way of updating options
+// without branking input in react-select v0.6.x
```
FrancescoCioria: typo ("branking")

src/time-picker/TimePicker.js
```diff
@@ -0,0 +1,226 @@
...
+  getLocals() {
+    const {
+      timeFormat,
+      className,
+      value: userValue,
+      ...props
+    } = this.props;
+
+    const value = userValue ? toOption({
+      ...userValue, timeFormat: this.props.timeFormat
```
FrancescoCioria: you already has the variable `timeFormat` in scope. You can safely replace `timeFormat: this.props.timeFormat` with just `timeFormat`

PR #315, opened 2016-02-22
#314: linkState should patch value (closes #314)

### LINKED ISSUE #314: linkState should patch value
## requirements
when doing a deep set `linkState` should patch value instead of replacing it

## specs
`set({...prevValue}, path, value)`

FrancescoCioria:
Issue #314


francescogior:
:octopus: :+1:

PR #319, opened 2016-02-24
#318: 'Add paths to icon with multipath' (closes #318)

### LINKED ISSUE #318: 'Add paths to icon with multipath'
See https://github.omnilab.our.buildo.io/buildo/labonline/issues/1522#issuecomment-4983

francescogior:
Issue #318

/cc @ascariandrea

### DIFF
--- a/src/Icon/Icon.js
+++ b/src/Icon/Icon.js
@@ -30,7 +30,7 @@ export default class Icon extends React.Component {
   template({ icon, className, style, onClick, paths }) {
     return icon ? (
       <i className={cx('icon', `icon-${icon}`, className)} style={style} onClick={onClick}>
-        {paths > 1 && range(paths).map(k => <span className={`path${k+1}`} /> )}
+        {paths > 1 && range(paths).map(k => <span className={`path${k + 1}`} /> )}
       </i>
     )
     : null;
--- a/src/Icon/Icon.js
+++ b/src/Icon/Icon.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import range from 'lodash/utility/range';
 import { pure, skinnable, props, t } from '../utils';
 import cx from 'classnames';
 
@@ -8,12 +9,14 @@ import cx from 'classnames';
   icon: t.maybe(t.Str),
   className: t.maybe(t.Str),
   style: t.maybe(t.Obj),
+  paths: t.maybe(t.refinements(t.Number, x => x - Math.floor(x) === 0 && x > 0)),
   onClick: t.maybe(t.Func)
 })
 export default class Icon extends React.Component {
 
   static defaultProps = {
     className: '',
+    paths: 1,
     style: {},
     onClick: () => {}
   };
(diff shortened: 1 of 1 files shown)


src/Icon/Icon.js
```diff
@@ -8,12 +9,14 @@ import cx from 'classnames';
   icon: t.maybe(t.Str),
   className: t.maybe(t.Str),
   style: t.maybe(t.Obj),
+  paths: t.maybe(t.refinements(t.Number, x => x - Math.floor(x) === 0 && x > 0)),
```
FrancescoCioria: please define `const Integer = t.refinement(t.Number, (n) => n % 1 === 0, 'Integer');` (copy-pasted from `tcomb` examples)
and then `paths: t.maybe(Integer)`

src/Icon/Icon.js
```diff
@@ -8,12 +9,14 @@ import cx from 'classnames';
...
   onClick: t.maybe(t.Func)
 })
 export default class Icon extends React.Component {
 
   static defaultProps = {
     className: '',
```
FrancescoCioria: as you're already managing this component, please remove `style` and `className` from `defaultProps`

src/Icon/Icon.js
```diff
@@ -24,9 +27,14 @@ export default class Icon extends React.Component {
     };
   }
 
-  template({ icon, className, style, onClick }) {
-    return icon ?
-      <i className={cx('icon', `icon-${icon}`, className)} style={style} onClick={onClick}/>
+  template({ icon, className, style, onClick, paths }) {
+    return icon ? (
+      <i className={cx('icon', `icon-${icon}`, className)} style={style} onClick={onClick}>
+        {paths === 1 ? null : range(paths).map(k => (
```
FrancescoCioria: replace with `paths > 1 &&` ?

francescogior: I grew up knowing that `React` doesn't like `false` as `children` but if you tell me that's ok I trust yoy

src/Icon/Icon.js
francescogior: :+1:  but 
`const PositiveInteger = t.refinement(t.Number, (n) => n % 1 === 0 && n > 0, 'PositiveInteger');`

src/Icon/Icon.js
francescogior: ok for `className` that is given to `cx` (`''` and `undefined` are treated the same), but what about style={null} ?

ascariandrea:
Maybe create a new specific component for dealing with `icomoon` icons? 
Cause having a general `Icon` component that renders *n* span with `className={`path${x}`}` is too opinionated?

francescogior:
added fixes, please @ascariandrea try this out

francescogior:
you're right, but I think this was built already quite 'ad hoc' for `icomoon`

FrancescoCioria:
> you're right, but I think this was built already quite 'ad hoc' for icomoon

yes, `Icon` currently works *only* for `icomoon`

FrancescoCioria:
CI: `33:70  error  Infix operators must be spaced  space-infix-ops`

ascariandrea:
Maybe we need a better name though..
I'm going to try it in the afternoon!

ascariandrea:
Tested and it works well, for me it can be merged!

PR #325, opened 2016-03-03
#266: [BackgroundDimmer] (closes #266)

### LINKED ISSUE #266: [BackgroundDimmer]
## description
when `stopScrollPropagation={true}` scroll is blocked everywhere, also inside the content.

FrancescoCioria:
Issue #266

### DIFF
--- a/examples/components/background-dimmer/BackgroundDimmer.example
+++ b/examples/components/background-dimmer/BackgroundDimmer.example
@@ -1,32 +1,23 @@
-const Example = React.createClass({
+class Example extends React.Component {
 
-  propTypes: {},
+  constructor() {
+  	super();
+    this.state = {};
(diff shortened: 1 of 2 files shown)


FrancescoCioria:
@giogonzo I dropped support for the (buggy and dangerous) stop of scrollPropagation when scrolling inside the `children`. Now It's up to the user stopping that event which can be easily done with a `ScrollView`.

The reason I'm not wrapping `children` in a `ScrollView` with `scrollPropagation={false}` is that this wouldn't work if the user passes another `ScrollView` (or simply a scrolling `div`) as `children`.

FrancescoCioria:
@giogonzo fixed CI errors

giogonzo:
nice!

PR #328, opened 2016-03-04
#308: [Accordion] rename in Section (closes #308)

### LINKED ISSUE #308: [Accordion] rename in Section
## requirements
The current definition of `Accordion` is different from the standard one from Bootstrap, Semantic-UI...
We should rename it into `Section`

FrancescoCioria:
Issue #308


francescogior:
:cactus:

PR #337, opened 2016-03-08
#336: [TimePicker] better sort for options in H12 format (closes #336)

### LINKED ISSUE #336: [TimePicker] better sort for options in H12 format
## description
right now `12:xx am` and `12:xx pm` appear as first result. They should be the last ones.
image

## specs
add a final sorting step to `makeFilterOptions`

nerder:
Issue #336

### DIFF
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -133,7 +133,7 @@ export const parseInTimeFormat = (inputStr, timeFormat) => {
 
 export const createTimeList = ({ hours, minutes }, timeFormat) => {
   if (!isValidHoursInTimeFormat(hours, timeFormat) || !Minute.is(minutes)) {
-    const hoursList = range(0, 24);
+    const hoursList = range(1, 24).concat(0);
     const minutesList = range(0, 60, interval);
     return flatten(hoursList.map(hours => minutesList.map(minutes => ({ hours, minutes, timeFormat }))));
   } else {
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import Dropdown from '../dropdown/Dropdown';
 import range from 'lodash/utility/range';
 import flatten from 'lodash/array/flatten';
+import sortBy from 'lodash/collection/sortBy';
 
 const options = [];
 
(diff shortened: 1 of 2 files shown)


src/time-picker/TimePicker.js
```diff
@@ -155,6 +156,8 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
   return lteTime(minTime, time) && lteTime(time, maxTime) && startsWith(time, originalInput);
 };
 
+export const sortTimeList = timeList => sortBy(timeList, time => ( time.hours === 0 ? 24 : time.hours));
```
FrancescoCioria: aren't you breaking the order of H24 options this way?
Also, why are you exporting it? If you plan to improve tests you can:
- do it directly in this PR
- open an issue

nerder: > aren't you breaking the order of H24 options this way?
No, because H24 goes from 0 to 23.
> Also, why are you exporting it?
Yes is because i want to improve tests, ill do here

FrancescoCioria: image

nerder: how did you do that? I don't get this behaviour in my examples

image

FrancescoCioria: yes, because you have a `minTime` so you never see the case `00:xx`

nerder: But this is concerning another issue is not directly related to this one, and is a design one. (that is actually scary, and hopefully you found it)

FrancescoCioria: I'm confused... why can't you simply refactor that function as:
```js
const sortTimeList = (timeList, timeFormat) => sortBy(timeList, time => (timeFormat === H12 && time.hours === 0 ? 24 : time.hours));
```

nerder:
After a discussion offline with @FrancescoCioria and after finding out some evidence that i will report down here, i can say that `times` as they are expressed don't have to be sorted. Even when they are in 12h format is actually correct to have 12:00 am as first option, for the exact same reason why 00:00 will be the first in 24h format. As you can see in this table:
image

After a small research on other `TimePicker` we can tell that actually this is the implementative choice that all of them have actually taken:

- http://timepicker.co/options/
- http://jonthornton.github.io/jquery-timepicker/

FrancescoCioria:
agreed. closing

PR #338, opened 2016-03-08
#324: [dropdown] drop support for `children` (closes #324)

### LINKED ISSUE #324: [dropdown] drop support for `children`
## requirements
when passing options as `children` nodes you loose the ability to set custom labels/values. This may cause bugs/defects and it doesn't add any useful functionality (this feature doesn't exists in `react-select`). We should drop support for `children`

## specs
remove `children` prop
throw warning if user passes `children`

nerder:
Issue #324


nerder:
The issue is resolved, but i can tell that the implementation of this component is very old.
Would you open another issue to refactor this code and add some tests? 

cc @FrancescoCioria @giogonzo

src/dropdown/Dropdown.js
```diff
@@ -20,34 +20,24 @@ const PropTypes = {
...
+  getGeneralProps = () => omit(this.props, Object.keys(PropTypes));
 
-      return {
-        options,
-        valueRenderer: this.renderValue,
-        optionRenderer: this.renderOption
-      };
-    }
+  logWarnings = () => {
+    warn(() => {
```
FrancescoCioria: in this case it's easier (and more readable) to write
```js
if (...) {
  warn('You\'re ... ');
}
```

src/dropdown/Dropdown.js
```diff
@@ -20,34 +20,24 @@ const PropTypes = {
 @props(PropTypes, { strict: false })
 export default class Dropdown extends React.Component {
 
+  componentDidMount() {
+    this.logWarnings();
+  }
+
   getChildren = () => [].concat(this.props.children || []);
```
FrancescoCioria: this should be removed

src/dropdown/Dropdown.js
```diff
@@ -20,34 +20,24 @@ const PropTypes = {
...
 
   renderOption = (option) => this.getChildren()[option.value];
```
FrancescoCioria: this should be removed

src/dropdown/Dropdown.js
```diff
@@ -20,34 +20,24 @@ const PropTypes = {
...
 
   renderValue = (option) => this.getChildren()[option.value];
```
FrancescoCioria: this should be removed

src/dropdown/Dropdown.js
nerder: ok, i will just have to revert this last commit then. I did it in that way at first, but i wasn't sure

FrancescoCioria: :+1:

FrancescoCioria:
Great!
Before merging can you please remove this (now) wrong example too? https://github.com/buildo/react-components/blob/master/examples/components/Dropdown/DropdownWithChildren.example

(remember to fix `examples.json` too)

FrancescoCioria:
great

PR #340, opened 2016-03-08
#306: [FlexView]  if basis='auto' throw warning (closes #306)

### LINKED ISSUE #306: [FlexView]  if basis='auto' throw warning
## requirements
`basis: 'auto'` is the default. If you pass `basis='auto'` we don't want to change the default for `shrink` to `shrink: false` -> we should keep `shrink: true` as default and throw a warning.

## specs
- if `basis === 'auto'` warn to remove and that `shrink` will still default to `true`
- if `basis === 'auto'` keep standard default for `shrink`

nerder:
Issue #306

### DIFF
--- a/src/flex/FlexView.js
+++ b/src/flex/FlexView.js
@@ -74,7 +74,7 @@ export default class FlexView extends React.Component {
 
   logWarnings = () => {
     if (this.props.basis === 'auto') {
-      warn('Default basis is auto. Force it to auto will leave shrink true as default');
+      warn('basis is "auto" by default: forcing it to "auto"  will leave "shrink:true" as default');
     }
   };
 
--- a/src/flex/FlexView.js
+++ b/src/flex/FlexView.js
@@ -73,11 +73,9 @@ export default class FlexView extends React.Component {
   }
 
   logWarnings = () => {
-    warn(() => {
-      if (this.props.basis === 'auto') {
-        return 'Default basis is auto. Force it to auto will leave shrink true as default';
-      }
-    });
+    if (this.props.basis === 'auto') {
+      warn('Default basis is auto. Force it to auto will leave shrink true as default');
+    }
   };
 
   getGrow = () => {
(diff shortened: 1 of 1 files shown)


src/flex/FlexView.js
```diff
@@ -68,6 +68,18 @@ const PropTypes = {
 @props(PropTypes, { strict: false })
 export default class FlexView extends React.Component {
 
+  componentDidMount() {
+    this.logWarnings();
+  }
+
+  logWarnings = () => {
+    warn(() => {
```
FrancescoCioria: same thing as other PR ( #338 )

nerder: ok maybe i don't get when is useful to use `warn` in that way. Fix it asap

src/flex/FlexView.js
```diff
@@ -68,6 +68,18 @@ const PropTypes = {
...
+      if (this.props.basis === 'auto') {
+        return 'Default basis is auto. Force it to auto will leave shrink true as default';
```
FrancescoCioria: I would change first part of the sentence to
basis is "auto" by default: forcing it to "auto" ...
(I think we are already using double quotes for props in other components)

src/flex/FlexView.js
```diff
@@ -68,6 +68,16 @@ const PropTypes = {
 @props(PropTypes, { strict: false })
 export default class FlexView extends React.Component {
 
+  componentDidMount() {
+    this.logWarnings();
+  }
+
+  logWarnings = () => {
+    if (this.props.basis === 'auto') {
+      warn('Default basis is auto. Force it to auto will leave shrink true as default');
```
FrancescoCioria: "Force it to auto" is an error in english :stuck_out_tongue_closed_eyes: 
anyway, I would change the first part to `basis is "auto" by default: forcing it to "auto" ....`
(we're already using double quotes for props in other warnings I think)

FrancescoCioria:
:taxi:

PR #341, opened 2016-03-08
#107: [TextOverflow] Text ellipsis but no text ellipsis method called (closes #107)

### LINKED ISSUE #107: [TextOverflow] Text ellipsis but no text ellipsis method called
We experienced a bug on Labonline where setting a width of a box with TextOverflow component inside leads to an ellipsis of text, but not the popover shown.

~~https://github.com/buildo/react-components/blob/master/src/text-overflow/TextOverflow.js#L35~~ (No more valid line)

Maybe this should change and check for minor and **equal** `offsetWidth`.

nerder:
Issue #107


FrancescoCioria:
:cake:

Issue #108, opened 2015-11-19
[Dropdown] defects


francescogior:
## Requirements
- When text input is dirty and gets in `blur` state, the `Dropdown` keeps displaying the value in the input. It should instead display the current selected option (or the placeholder 'Select...')
- When in focus, selected option or placeholder should have reduced opacity (placeholderish) to induce the user to understand he can type text to search options

nerder:
@francescogior I'm going to fix this issue but i have some question about the second bullet point. If i understand well, you want to reduce opacity on the label of selected option. I don't very agree with this, because it can be sound confusing for the users. No real evidence of why, it just seems weird to me.

CC @FrancescoCioria

francescogior:
look here!

[![https://gyazo.com/6ce83519af6fd2d47957a59d9c2dda7d](https://i.gyazo.com/6ce83519af6fd2d47957a59d9c2dda7d.gif)](https://gyazo.com/6ce83519af6fd2d47957a59d9c2dda7d)

nerder:
Ok i got it. I can split this in 2 different issues closing this one?

francescogior:
I think it's ok to work on this one for both the two tasks!

FrancescoCioria:
About the second point I remind you that this is the default style (we're on `react-components`) --> It's likely to be overwritten in any project.

@francescogior why do you think it would useful a different default? wouldn't we change it anyway on our projects?

francescogior:
I think that when (and only when) the `Dropdown` is `searchable` the default style should be that one:
- if an options is selected, the `:focus` style of `Dropdown` should inform the users that they can type something. This is not clear with the current component in LOL
[![https://gyazo.com/345579344f15ec56fdefd016ac7577aa](https://i.gyazo.com/345579344f15ec56fdefd016ac7577aa.gif)](https://gyazo.com/345579344f15ec56fdefd016ac7577aa)
- It can be OK for this behaviour to be overridable but I don't see why one should do it!

nerder:
## Specs Proposal
Could i remove `valueLink` props from `DropDown`? It will not add anything and it will just add possible complications. (i.e: i have to repeat some checks for handle the same behaviour in both cases)

nerder:
With this pr i actually solve only the first bullet point (the `onBlur` reset value issue), in order to fix even the second bug i need this to be merged first #133 , i'll provide here the code that we have to use for fix the second point:

```css
.Select-control > .Select-placeholder {
  color: #333333;
[...]

PR #342, opened 2016-03-09
#133: [Dropdown] export react-select style from /dropdown (closes #133)

### LINKED ISSUE #133: [Dropdown] export react-select style from /dropdown
## requirements
- facilitate import of needed style
- make it more clear that this component comes with a `css` file
- follow our (soon to be) sass name convention: `camelCase` version of the component name (refer to point in #204 )

## specs
manually copy `node_modules/react-select/dist/default.css` to `dropdown/dropdown.scss`

[EDITED]

nerder:
Issue #133

### DIFF
--- a/examples/components/styles.js
+++ b/examples/components/styles.js
@@ -6,4 +6,3 @@ import '../../src/confirmation-input/confirmationInput.scss';
 import '../../src/badge/badge.scss';
 import '../../src/toggle/toggle.scss';
 import '../../src/table/table.scss';
-import '../../node_modules/react-select/dist/default.css';
--- a/examples/components/styles.js
+++ b/examples/components/styles.js
@@ -1,6 +1,7 @@
 import '../../src/more-or-less/moreOrLess.scss';
 import '../../src/flex/flexView.scss';
 import '../../src/loading-spinner/style.scss';
+import '../../src/dropdown/dropdown.scss';
 import '../../src/confirmation-input/confirmationInput.scss';
 import '../../src/badge/badge.scss';
 import '../../src/toggle/toggle.scss';
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -5,8 +5,6 @@ import find from 'lodash/find';
 import cx from 'classnames';
 import { warn } from '../utils/log';
 
-import './dropdown.scss';
-
 const PropTypes = {
   value: t.maybe(t.union([t.Number, t.String, t.Object])),
   valueLink: t.maybe(t.struct({
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -5,6 +5,8 @@ import find from 'lodash/find';
 import cx from 'classnames';
 import { warn } from '../utils/log';
 
+import './dropdown.scss';
+
 const PropTypes = {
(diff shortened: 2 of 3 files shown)


nerder:
NOTE: please before to merge this, make sure to close #324

src/dropdown/Dropdown.js
```diff
@@ -5,6 +5,8 @@ import cx from 'classnames';
 import { props, t } from '../utils';
 import { warn } from '../utils/log';
 
+import './dropdown.scss';
```
FrancescoCioria: This is a breaking change we don't want: every old project built without webpack is going to break if you add this import.
Please remove

src/dropdown/dropdown.scss
```diff
@@ -0,0 +1,276 @@
+/**
```
FrancescoCioria: this file should be named `style.scss`

src/dropdown/Dropdown.js
nerder: i saw this in other components, how am i suppose to import `style` then?

FrancescoCioria: you don't, the user will import it by himself (like we're currently doing)
ex:
```js
import 'buildo-react-components/src/dropdown/style.scss'
```

src/dropdown/dropdown.scss
FrancescoCioria: wait

FrancescoCioria: I updated the specs: leave as is.

FrancescoCioria:
@nerder why does it have so many commits now? this morning it had only the last two (the correct ones)

nerder:
i mess up with rebase i guess, need to cleanup

nerder:
Can be merged now @FrancescoCioria

src/dropdown/Dropdown.js
```diff
@@ -5,6 +5,8 @@ import find from 'lodash/find';
 import cx from 'classnames';
 import { warn } from '../utils/log';
 
+import './dropdown.scss';
```
FrancescoCioria: why is this still here? please remove

examples/components/styles.js
```diff
@@ -1,6 +1,7 @@
 import '../../src/more-or-less/moreOrLess.scss';
 import '../../src/flex/flexView.scss';
 import '../../src/loading-spinner/style.scss';
+import '../../src/dropdown/dropdown.scss';
 import '../../src/confirmation-input/confirmationInput.scss';
 import '../../src/badge/badge.scss';
 import '../../src/toggle/toggle.scss';
```
FrancescoCioria: you should remove import of the old style: `import '../../node_modules/react-select/dist/default.css';`
(it's two lines below but I can't comment on it)

nerder: Completely didn't seen. Thank you for that.

PS: you want me to add here the css mentioned in #108 on in another pr?

FrancescoCioria: better in another PR and it should be added to the main style (`dropdown.scss`)

nerder: ok i'll open another issue then

FrancescoCioria:
:vhs:

nerder:
:hand:

PR #343, opened 2016-03-09
#108: [Dropdown] defects (closes #108)

### LINKED ISSUE #108: [Dropdown] defects

nerder:
Issue #108


examples/components/Dropdown/default.example
```diff
@@ -7,6 +12,7 @@ class Example extends React.Component {
       { value: 'orange', label: 'Orange' },
       { value: 'banana', label: 'Banana' }
     ];
+    //const valueLink = { value: this.state.value, requestChange: value => this.setState({ value }) };
```
FrancescoCioria: remove?

src/dropdown/Dropdown.js
```diff
@@ -1,22 +1,20 @@
...
 import { warn } from '../utils/log';
 
-const themes = {
-  semantic: 'semantic-theme'
-};
-
 const PropTypes = {
-  children: t.maybe(t.ReactNode),
-  theme: t.maybe(t.enums.of(['semantic'])),
-  valueLink: t.maybe(t.struct({
```
FrancescoCioria: why did you remove `valueLink`? please revert this

src/dropdown/Dropdown.js
```diff
@@ -24,34 +22,46 @@ export default class Dropdown extends React.Component {
...
-  render() {
-    // The order is important: props may override previous ones
-    return (
-      <Select
-        {...this.getGeneralProps()}
-        {...this.getValueLinkProps()}
-        {...this.getClassName()}
-      />
-    );
+  _onChange = () => this.props.valueLink ? this.props.valueLink.requestChange : this.props.onChange;
```
FrancescoCioria: 1. `_onChange` is called with a `value`
2. this is an usual why of defining a `_onChange` function: why aren't you calling the prop `onChange` here?
I think you should do this:
```js
_onChange(value) => this.props.valueLink ? this.props.valueLink.requestChange(value) : this.props.onChange(value);
```

or if you prefer your version then please rename it to `getOnChange` (as that's what you're actually doing)
this.props.valueLink ? this.props.valueLink.requestChange() : this.props.onChange;

src/dropdown/Dropdown.js
```diff
@@ -24,34 +22,46 @@ export default class Dropdown extends React.Component {
...
+      valueLink,
+      ...props
+    } = this.props;
+
+    return {
+      ...props,
+      options,
+      className: cx('dropdown', className),
+      value: this.valueToOption(this.getValue(), options),
+      onChange: this._onChange(),
```
FrancescoCioria: as pointed in the comment above this should be either `this._onChange` or `this.getOnChange()`

src/dropdown/Dropdown.js
nerder: actually i prefer my version that return the function.
I should rename it then

src/dropdown/Dropdown.js
```diff
@@ -1,22 +1,24 @@
...
-const themes = {
-  semantic: 'semantic-theme'
-};
-
 const PropTypes = {
-  children: t.maybe(t.ReactNode),
-  theme: t.maybe(t.enums.of(['semantic'])),
+  value: t.maybe(t.union([t.Number, t.String, t.Object])),
   valueLink: t.maybe(t.struct({
     value: t.maybe(t.String),
```
FrancescoCioria: as you improved the evaluation for `value`, can you improve this too?
(`value: t.maybe(t.union([t.Number, t.String, t.Object])),`)

FrancescoCioria:
:kimono:

PR #347, opened 2016-03-10
#344: [Dropdown] Add Unit Test (closes #344)

### LINKED ISSUE #344: [Dropdown] Add Unit Test
## requirements
Add some tests to this component

## specs
Now that we have `skinnable()` is way to easy to test this component
## misc
{optional: other useful info}

nerder:
Issue #344

### DIFF
--- a/test/tests/components/Dropdown.js
+++ b/test/tests/components/Dropdown.js
@@ -1,0 +1,29 @@
+import expect from 'expect';
+import Dropdown, { Props } from '../../../src/dropdown';
+import { newComponent } from '../helpers';
+
+describe('Dropdown', function () {
+
+  describe('locals', () => {
+    const exampleProps = {
+      id: '12345',
+      className: 'fancy-class-name',
+      style: { margin: 10, position: 'relative' }
+    };
+    const componentDropdown = new Dropdown(exampleProps);
+
+    it('should pass props', ()  => {
+      const locals = componentDropdown.getLocals();
+      expect(locals.id).toBe(exampleProps.id);
+      expect(locals.style).toEqual(exampleProps.style);
+    });
+
+    it('should compute className', () => {
+      const { className } = componentDropdown.getLocals();
+      expect(className).toInclude('dropdown');
+      expect(className).toInclude('fancy-class-name');
+    });
+
+  });
+
+});
--- a/test/tests/components/Dropdown.js
+++ b/test/tests/components/Dropdown.js
@@ -1,15 +1,15 @@
 import expect from 'expect';
-import Dropdown, { Props } from '../../../src/dropdown';
+import Dropdown from '../../../src/dropdown';
 import { newComponent } from '../helpers';
 
-describe('Dropdown', function () {
+describe('Dropdown', () => {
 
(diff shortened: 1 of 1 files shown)


test/tests/components/Dropdown.js
```diff
@@ -0,0 +1,72 @@
+import expect from 'expect';
+import Dropdown, { Props } from '../../../src/dropdown';
+import { newComponent } from '../helpers';
+
+describe('Dropdown', function () {
```
FrancescoCioria: you should use arrow functions

test/tests/components/Dropdown.js
```diff
@@ -0,0 +1,72 @@
...
+      expect(className).toInclude('fancy-class-name');
+    });
+
+    it('should compute value', () => {
+      const { value } = componentDropdown.getLocals();
+      expect(value).toBeA(Object);
+      expect(value).toEqual(exampleProps.value);
+    });
+
+    it('should compute value as string', () => {
```
FrancescoCioria: `should compute value **from** a string` :)

test/tests/components/Dropdown.js
```diff
@@ -0,0 +1,72 @@
...
+      expect(locals.onChange).toEqual(exampleProps.onChange);
+    });
+
+    it('should compute className', () => {
+      const { className } = componentDropdown.getLocals();
+      expect(className).toInclude('dropdown');
+      expect(className).toInclude('fancy-class-name');
+    });
+
+    it('should compute value', () => {
```
FrancescoCioria: `should leave object value untouched` (we're testing that if you pass `value` as an object it remains untouched)

test/tests/components/Dropdown.js
```diff
@@ -0,0 +1,72 @@
...
+import { newComponent } from '../helpers';
+
+describe('Dropdown', function () {
+
+  describe('locals', () => {
+    const exampleProps = {
+      id: '12345',
+      className: 'fancy-class-name',
+      style: { margin: 10, position: 'relative' },
+      value: { value: 'test', label: 'Test' },
```
FrancescoCioria: I think you should add other custom properties to this object. We're testing that `Dropdown` doesn't touch a `value` object and that should be true for **any** object even if malformed.

FrancescoCioria:
Your tests look ok (apart from a couple of descriptions :stuck_out_tongue_closed_eyes: ) the problem I see is that you're only testing that everything works with valid inputs. It's useful (sometimes even more useful) to test also the opposite:
- what should happen if `Dropdown` receives an empty object as `value`?
- what should happen if `Dropdown` receives an `array` as `value`?
- what should happen if `Dropdown` receives a string as `options`?

or also lighter cases like
- what should happen if `Dropdown` receives an array of options malformed? (with different/more properties than just `value` and `label`)

By doing those tests too you force yourself to think about corner cases and therefore to find a solution (or at least to define a behavior) for many more cases than just the three correct ones.

Ex: you could make a test where you pass a wrong `value` (`value: 'unreal_id') that you can't find in the `options`. What should happen in this case? should you throw a warning? maybe yes

PR #348, opened 2016-03-10
#346: [TextOverflow] - treats not overflowed text as such (closes #346)

### LINKED ISSUE #346: [TextOverflow] - treats not overflowed text as such
## description
- this one has correctly a tooltip

[https://gyazo.com/6b5a2603efe6ae16ed7b105a6eb799a8](https://gyazo.com/6b5a2603efe6ae16ed7b105a6eb799a8)

- but this one shouldn't

[https://gyazo.com/05866011d6b8e63310bc6471b7cf05fd](https://gyazo.com/05866011d6b8e63310bc6471b7cf05fd)

FrancescoCioria:
Issue #346

### DIFF
--- a/src/text-overflow/TextOverflow.js
+++ b/src/text-overflow/TextOverflow.js
@@ -49,9 +49,8 @@ export default class TextOverflow extends React.Component {
     });
   };
 
-  verifyOverflow = (_state) => {
-    const state = _state || this.state;
-    if (state.isOverflowing === false && window) {
+  verifyOverflow = () => {
+    if (this.state.isOverflowing === false && window) {
       const text = React.findDOMNode(this.refs.text);
       const textWithoutEllipsis = React.findDOMNode(this.refs.textWithoutEllipsis);
 
--- a/src/text-overflow/TextOverflow.js
+++ b/src/text-overflow/TextOverflow.js
@@ -40,7 +40,7 @@ export default class TextOverflow extends React.Component {
 
   logWarnings = () => {
     warn(() => {
-      const node = React.findDOMNode(this.refs.text).parentNode;
+      const node = React.findDOMNode(this.refs.text).parentNode.parentNode;
       const { width, flex } = node.style;
       const flexBasis = flex ? flex.split(' ')[2] : null;
       if (width !== '100%' && flexBasis !== '100%') {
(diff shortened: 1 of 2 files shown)


FrancescoCioria:
@giogonzo I'm rendering an invisible `span` with the same content and compare it with the first one to understand if it's overflowing or not... (Fuck the system!)

I'm using `element.offsetWidth` instead of `parseFloat(window.getComputedStyle(element).width)` because the browser (in corner case situations) may render the first one with a smaller width but **without** ellipsis: if we have to live with this inconsistency I'd rather use the easier (and isomorphic friendly) way

src/text-overflow/TextOverflow.js
```diff
@@ -26,9 +26,9 @@ export default class TextOverflow extends React.Component {
     this.state = { isOverflowing: false };
   }
 
-  componentDidMount = () => this.verifyOverflow();
+  componentDidMount = () => setTimeout(this.verifyOverflow);
```
giogonzo: why exactly the `setTimeout`s here?

src/text-overflow/TextOverflow.js
```diff
@@ -51,9 +51,15 @@ export default class TextOverflow extends React.Component {
 
   verifyOverflow = (_state) => {
```
giogonzo: is the `_state` argument ever passed?

src/text-overflow/TextOverflow.js
```diff
@@ -51,9 +51,15 @@ export default class TextOverflow extends React.Component {
...
     const state = _state || this.state;
-    if (state.isOverflowing === false) {
+    if (state.isOverflowing === false && window) {
```
giogonzo: `typeof window !== 'undefined'` is nicer, this would be a reference error in `node`

src/text-overflow/TextOverflow.js
FrancescoCioria: apparently not, maybe it was used like in an old version. I'll fix it

src/text-overflow/TextOverflow.js
FrancescoCioria: discussed offline (`window.getComputedStyle(el)` returns a wrong value without `setTimeout`)

giogonzo:
ok let's try this!

PR #350, opened 2016-03-10
#349: [Dropdown] Selected option reduce opacity (closes #349)

### LINKED ISSUE #349: [Dropdown] Selected option reduce opacity
## requirements
When on focus, selected option or placeholder should have reduced opacity (placeholderish) to induce the user to understand he can type text to search options

## specs
you should edit the file ´dropdown.scss´ with the code posted here #108

nerder:
Issue #349

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -22,19 +22,6 @@
   padding: 8px 52px 8px 10px;
 }
 
-.Select-control > .Select-placeholder {
-  color: #333333;
-  opacity: .5;
-}
-
-.has-value > .Select-control > .Select-placeholder {
-  opacity: 1;
-}
-
-.Select.dropdown.is-searchable.is-open.is-focused .Select-placeholder {
-  opacity: .75;
-}
-
 .Select-control:hover {
   box-shadow: 0 1px 0 rgba(0, 0, 0, 0.06);
 }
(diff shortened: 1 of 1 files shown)


src/dropdown/dropdown.scss
```diff
@@ -21,9 +21,24 @@
...
+.Select-control > .Select-placeholder {
+  color: #333333;
+  opacity: .5;
+}
+
+.has-value > .Select-control > .Select-placeholder {
+  opacity: 1;
+}
+
+.Select.dropdown.is-searchable.is-open.is-focused .Select-placeholder {
```
FrancescoCioria: this one looks ok

src/dropdown/dropdown.scss
```diff
@@ -21,9 +21,24 @@
   outline: none;
   padding: 8px 52px 8px 10px;
 }
+
+.Select-control > .Select-placeholder {
```
FrancescoCioria:
```scss
.has-value {
  > .Select-control {
    > .Select-placeholder {
      ...
    }
  }

**Comment by FrancescoCioria** [comment]
@nerder as the original file was a `.css` all of it is written in plain `css`. I say we forget about it and use `.scss` syntax for our additions (as you already meant to do) but, to make it more explicit, maybe we should add a comment line to separate our custom style from the default (something like `// CUSTOM STYLE`)

File: src/dropdown/dropdown.scss (hunk 2 above)
Comment (nerder) [inline src/dropdown/dropdown.scss]: you mean:

```
.has-value {
  .Select-control {
    .Select-placeholder {
      opacity: 1;
    }
  }
}
```

File: src/dropdown/dropdown.scss [hunk 3]
```diff
@@ -21,9 +21,11 @@
   outline: none;
   padding: 8px 52px 8px 10px;
 }
+
```
FrancescoCioria: why did you add these two empty lines? (only here?)
please remove :P

PR #352, opened 2016-03-15
#351: [background-dimmer] content should scroll when too big (closes #351)

### LINKED ISSUE #351: [background-dimmer] content should scroll when too big
## description
when content is higher/wider the the window the layout breaks. Instead of overflowing the window the content should scroll

## specs
- give content wrapper `div` a `className`
- make it a `FlexView` column
- add an arbitrary default `maxHeight` and `maxWidth` (I say between 80/90 %)

FrancescoCioria:
Issue #351

image

- if content is higher/wider than 90% of window it correctly scrolls! (if `children` has `overflow: scroll` of course :dancer: )

### DIFF
--- a/src/background-dimmer/BackgroundDimmer.js
+++ b/src/background-dimmer/BackgroundDimmer.js
@@ -46,7 +46,7 @@ export default class BackgroundDimmer extends React.Component {
 
   isEventOutsideChildren = (e) => {
     const el = e.target || e.srcElement;
-    return el === React.findDOMNode(this.refs.flexWrapper);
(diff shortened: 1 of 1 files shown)


giogonzo:
:+1:  from what I can tell

the arbitrary `90%` is weird?

FrancescoCioria:
yeah, I thought about exposing the props but I decided not to because I meant to put it in the sass...only that there's no sass :) I'm going to add two props to make it easily overridable

giogonzo:
:+1:

PR #354, opened 2016-03-15
#334: Popover is not shown/open when `popover.isOpen` starts `true` (closes #334)

### LINKED ISSUE #334: Popover is not shown/open when `popover.isOpen` starts `true`
## description

rendering `Popover` `isOpen=true` doesn't show the popover as expected

## specs
In `componentDidMount`, if `this.isOpen()` is true, we should call `this.onPopoverOpenChange` which is already being used anywhere else in code.
(replace this: https://github.com/buildo/react-components/blob/master/src/popover/Popover.js#L54)

EDIT by Francesco
## misc

added an example here that exposes the defect behavior: https://github.com/buildo/react-components/tree/popover-initially-open

FrancescoCioria:
Issue #334

### DIFF
--- a/examples/components/examples.json
+++ b/examples/components/examples.json
@@ -161,7 +161,8 @@
       "component": "react-components/__TAG__/src/popover/Popover.js",
       "examples": [
         "react-components/__TAG__/examples/components/popover/Relative.example",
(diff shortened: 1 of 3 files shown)


giogonzo:
lgtm :+1:

PR #357, opened 2016-03-16
#356: [TextOverflow] Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'. (closes #356)

### LINKED ISSUE #356: [TextOverflow] Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
## description

under certain circumstances, TextOverflow is trying to get the computed style of a `null` `React.findDOMNode` return value:
```
Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
```

## specs

for the time being, just patch it and hide the possibly-null access behind a check

giogonzo:
Issue #356

### DIFF
--- a/src/text-overflow/TextOverflow.js
+++ b/src/text-overflow/TextOverflow.js
@@ -40,11 +40,14 @@ export default class TextOverflow extends React.Component {
 
   logWarnings = () => {
     warn(() => {
-      const node = React.findDOMNode(this.refs.text).parentNode.parentNode;
-      const { width, flex } = node.style;
-      const flexBasis = flex ? flex.split(' ')[2] : null;
-      if (width !== '100%' && flexBasis !== '100%') {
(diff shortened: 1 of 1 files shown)


src/text-overflow/TextOverflow.js
```diff
@@ -41,10 +41,12 @@ export default class TextOverflow extends React.Component {
   logWarnings = () => {
     warn(() => {
       const node = React.findDOMNode(this.refs.text).parentNode.parentNode;
```
FrancescoCioria: what if `React.findDOMNode(this.refs.text)` is `null`? Are we sure this could never happen?

giogonzo: sure, didn't notice the `.parent..`, fixing now

giogonzo:
@FrancescoCioria should be "ok" now

FrancescoCioria:
:+1:

PR #359, opened 2016-03-16
#358: [Table] Add prop to handle column resize (closes #358)

### LINKED ISSUE #358: [Table] Add prop to handle column resize
## requirements
- Table should provide the same `fixed-data-table` api to deal with column resizing

## specs
- implement `onColumnResizeEndCallback: t.maybe(t.Function)` and `isColumnResizing: t.maybe(t.Boolean)` as props

ascariandrea:
Issue #358

## Test Plan
- tested on `LOL` and it works
 - i was able to set column width on `onColumnResizeEndCallback` invocation
 - i was able to hide black vertical line by passing `isColumnResizing = false` on `onColumnResizeEndCallback` call.


francescogior:
:+1:  :hurtrealbad:

PR #361, opened 2016-03-22
#360: [TabbedPanel] tabs should have same width (closes #360)

### LINKED ISSUE #360: [TabbedPanel] tabs should have same width
## description
The tabs should have the same width

## specs
replace `grow` with `basis='100%'` and `shrink`

nerder:
Issue #360

## Test Plan

- Open examples page
- Click on `Panel`
- The 3rd example will show you the expected behaviour

Note: if very very long, is not managed. It will be a very uncommon behaviour IMO, to me we can leave it free and make it editable by the user (with `overflow:hidden` or `text-overflow: ellipsis` for instance) 

cc @FrancescoCioria

### DIFF
--- a/examples/components/Panel/tabbed.example
+++ b/examples/components/Panel/tabbed.example
@@ -12,7 +12,7 @@ class Example extends React.Component {
     const panelProps = {
       type: 'floating',
       tabs: {
-        headers: [ 'Tab 1', 'Tab 2' ],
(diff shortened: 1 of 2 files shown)


FrancescoCioria:
> to me we can leave it free and make it editable by the user (with `overflow:hidden` or `text-overflow: ellipsis` for instance)

I agree, we shouldn't be concerned by this kind of corner cases.

The PR looks perfect, good job!

PR #365, opened 2016-03-30
#364: [Icon] Add color prop (closes #364)

### LINKED ISSUE #364: [Icon] Add color prop
## requirements
- it should be easier to give an icon a color

## specs
- add `color` prop
- it would override color defined in `style` prop

francescogior:
Issue #364

### DIFF
--- a/src/Icon/Icon.js
+++ b/src/Icon/Icon.js
@@ -9,6 +9,7 @@ const PositiveInteger = t.refinement(t.Number, x => x % 1 === 0 && x > 0, 'Posit
 @skinnable()
 @props({
   icon: t.maybe(t.Str),
+  color: t.maybe(t.String),
   className: t.maybe(t.Str),
   style: t.maybe(t.Obj),
   paths: t.maybe(PositiveInteger),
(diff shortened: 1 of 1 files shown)


src/Icon/Icon.js
```diff
@@ -27,9 +28,9 @@ export default class Icon extends React.Component {
     };
   }
 
-  template({ icon, className, style, onClick, paths }) {
+  template({ icon, color, className, style = {}, onClick, paths }) {
     return icon ? (
-      <i className={cx('icon', `icon-${icon}`, className)} style={style} onClick={onClick}>
+      <i className={cx('icon', `icon-${icon}`, className)} style={{ ...style, color: color || style.color }} onClick={onClick}>
```
FrancescoCioria: if you write it like this:
```jsx
style={{ ...style, color }}
```
does `color` overwrite `style.color` even when `color` is `undefined`?

francescogior: sadly, yes :(
I tried that way first :( :(

FrancescoCioria:
:gem:

PR #366, opened 2016-04-04
Fix width and max-width for BackgroundDimmer

ascariandrea:
- `maxWidth` shouldn't be a **relative** value but **fixed**
- and `width` should be define as `maxWidth` was defined before

### DIFF
--- a/src/background-dimmer/BackgroundDimmer.js
+++ b/src/background-dimmer/BackgroundDimmer.js
@@ -32,9 +32,17 @@ import FlexView from '../flex/FlexView';
    */
   onClickOutside: t.maybe(t.Function),
   /**
+   * centeredContentWrapper width
+   */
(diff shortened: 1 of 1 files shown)


FrancescoCioria:
@ascariandrea what is your need? this PR, as it is now, would break every modal in our projects as this would be the new default layout:
image

the content wrapper inside `BackgroundDimmer` must have the same size of its `children` and (if necessary) limit this size to a maximum (`max-width`, `max-height`).

If you set the `width` of content wrapper the the `children` must have `width: 100%` for it to be meaningful.
If you need this case you can:
1. overwrite `width` using `css`
2. modify this PR by:
  - setting `width: 'auto'` as default and leaving `maxWidth: '90%'` (which IMHO is a nice default)
  - adding `height` (we can't have different props for `height` and `width`)

FrancescoCioria:
perfect! :dart:

PR #372, opened 2016-04-12
#371: overridable panel css (closes #371)

### LINKED ISSUE #371: overridable panel css

francescogior:
Issue #371

### DIFF
--- a/src/Panel/PanelHeader.js
+++ b/src/Panel/PanelHeader.js
@@ -66,9 +66,9 @@ export default class PanelHeader extends React.Component {
 
   templateTitle = ({ renderTitle, title, renderExpandIcon, collapse }) => {
     return (
-      <FlexView vAlignContent='center' shrink={false} onClick={collapse && collapse.onToggleExpanded} className='panel-header-title-wrapper'>
+      <FlexView vAlignContent='center' shrink={false} onClick={renderExpandIcon && collapse.onToggleExpanded} className='panel-header-title-wrapper'>
         {renderExpandIcon && this.templateExpandIcon({ collapse })}
-        <FlexView className='panel-header-title'>{renderTitle && title}</FlexView>
+        {renderTitle && <FlexView className='panel-header-title'>{title}</FlexView>}
       </FlexView>
     );
 
--- a/src/Panel/PanelHeader.js
+++ b/src/Panel/PanelHeader.js
@@ -49,33 +49,27 @@ export default class PanelHeader extends React.Component {
     };
   }
 
-  templateExpandIcon = ({ renderExpandIcon, collapse }) => {
+  templateExpandIcon = ({ collapse }) => {
     return (
-      renderExpandIcon ?
-        <FlexView
-          vAlignContent='center'
-          hAlignContent='center'
-          style={{ cursor: 'pointer' }}
-          className='panel-header-icon'
-          basis={30}
-          onClick={collapse.onToggleExpanded}
-        >
-          <Icon
-            icon={this.getIcon(collapse)}
-            className='expand-icon'
-          />
-        </FlexView>
-        : null
+      <FlexView
+        vAlignContent='center'
+        hAlignContent='center'
+        className='panel-header-icon'
+      >
+        <Icon
+          icon={this.getIcon(collapse)}
+          className='expand-icon'
+        />
+      </FlexView>
     );
   };
 
-  templateTitle = ({ renderTitle, title }) => {
+  templateTitle = ({ renderTitle, title, renderExpandIcon, collapse }) => {
     return (
-      renderTitle ?
-        <FlexView vAlignContent='center' className='panel-header-title' shrink={false}>
-          {title}
-        </FlexView>
-        : null
+      <FlexView vAlignContent='center' shrink={false} onClick={collapse && collapse.onToggleExpanded} className='panel-header-title-wrapper'>
+        {renderExpandIcon && this.templateExpandIcon({ collapse })}
+        <FlexView className='panel-header-title'>{renderTitle && title}</FlexView>
+      </FlexView>
     );
 
   };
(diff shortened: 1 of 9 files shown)


src/Panel/PanelHeader.js
```diff
@@ -49,33 +49,27 @@ export default class PanelHeader extends React.Component {
...
-  templateTitle = ({ renderTitle, title }) => {
+  templateTitle = ({ renderTitle, title, renderExpandIcon, collapse }) => {
     return (
-      renderTitle ?
-        <FlexView vAlignContent='center' className='panel-header-title' shrink={false}>
-          {title}
-        </FlexView>
-        : null
+      <FlexView vAlignContent='center' shrink={false} onClick={collapse && collapse.onToggleExpanded} className='panel-header-title-wrapper'>
+        {renderExpandIcon && this.templateExpandIcon({ collapse })}
```
FrancescoCioria: what's the difference between `renderExpandIcon &&` and `collapse &&`? is `renderExpandIcon` a duplicate?

src/Panel/PanelHeader.js
```diff
@@ -49,33 +49,27 @@ export default class PanelHeader extends React.Component {
...
+        <FlexView className='panel-header-title'>{renderTitle && title}</FlexView>
```
FrancescoCioria: `renderTitle &&` should be outside the `FlexView`:
```jsx
renderTitle && <FlexView ...
```

src/Panel/panel.scss
```diff
@@ -1,4 +1,13 @@
-@import '../utils/theme/variables.scss';
+$border-color: #e6e6e6 !default;
+$soft-loader-color: #45a7b9 !default;
+$panel-header-background-color: #2c3a40 !default;
```
FrancescoCioria: please remove the prefix `panel-` from the variables as it's implicit (we're not adding a prefix in the others components)

src/Panel/panel.scss
```diff
@@ -1,4 +1,13 @@
...
+$panel-header-color: white !default;
+$expand-icon-color: white !default;
+$panel-header-title-font-weight: 700 !default;
+$panel-content-background-color: white !default;
+$panel-content-color: #2c3a40 !default;
+$panel-content-title-color: #2c3a40 !default;
+$panel-content-title-background-color:  white !default;
```
FrancescoCioria: why do we have this? shouldn't it be transparent (which is the default)?

src/Panel/panel.scss
```diff
@@ -54,42 +59,51 @@
...
+    > .panel-header .panel-header-title-wrapper {
+      cursor: pointer;
+    }
+  }
+
   > .panel-header {
-    color: $white;
-    background-color: $charcoal;
+    color: $panel-header-color;
+    background-color: $panel-header-background-color;
```
FrancescoCioria: when using a `background-` variable please use always `background: $...`. This is needed in case we have to customize not only the color (for example when using a gradient)

src/Panel/panel.scss
```diff
@@ -54,42 +59,51 @@
...
       font-size: 20px;
+      color: $expand-icon-color;
+      margin-right: 3px;
     }
 
     .panel-header-title {
       font-size: 16px;
-      font-weight: $semibold;
-      padding-left: 20px;
+      font-weight: $panel-header-title-font-weight;
```
FrancescoCioria: as discussed offline I personally wouldn't add a variable for `font-weight` but maybe in this case is useful

src/Panel/panel.scss
```diff
@@ -54,42 +59,51 @@
...
 
     .panel-menu {
       padding: 7px 0;
+      margin-right: -12px;
     }
   }
 
   > .panel-content {
-    background-color: $white;
+    background-color: $panel-content-background-color;
```
FrancescoCioria: same thing as above: please replace with `background: $content-background`

src/Panel/panel.scss
```diff
@@ -54,42 +59,51 @@
...
 
     .panel-content-title {
       font-size: 17px;
       margin: 0;
-      color: $charcoal;
-      background: $white;
+      color: $panel-content-title-color;
+      background: $panel-content-title-background-color;
```
FrancescoCioria: same

src/more-or-less/moreOrLess.scss
```diff
@@ -1,6 +1,7 @@
 $expand-button-background: #EAEAEA !default;
 $expand-button-hover-background: #F1F2F3 !default;
 $expand-button-icon-color: #999 !default;
+$expand-button-icon-hover-color: #999 !default;
```
FrancescoCioria: I'm following the convention of adding the state (`hover`, `open`, `has-value`....) always at the end of the variable. please replace with `$expand-button-icon-color-hover` :)

PS: sorry about this, I'll write a small readme after this wonderful design week

src/more-or-less/moreOrLess.scss
```diff
@@ -11,6 +12,9 @@ $expand-button-icon-color: #999 !default;
 
     &:hover {
       background-color: $expand-button-hover-background;
+      .expand-button-icon {
+        color: $expand-button-icon-hover-color;
```
FrancescoCioria: same thing, please move `-hover` at the end

src/Panel/PanelHeader.js
francescogior: It works the same but I think it's ok to keep it, just because it has an explicit name and explicit type (boolean)

src/Panel/PanelHeader.js
francescogior: :+1:

src/Panel/panel.scss
francescogior: :+1:

src/Panel/panel.scss
francescogior: It defined as white below

src/Panel/panel.scss
francescogior: :+1:

src/Panel/panel.scss
francescogior: :

francescogior:
:cactus: :cake: :panel:!

PR #391, opened 2016-04-16
#387: [button] make it compatible with new style guidelines (closes #387)

### LINKED ISSUE #387: [button] make it compatible with new style guidelines

francescogior:
Issue #387

### DIFF
--- a/src/button/Button.js
+++ b/src/button/Button.js
@@ -1,38 +1,52 @@
 import React from 'react';
 import cx from 'classnames';
 import pick from 'lodash/pick';
+import every from 'lodash/every';
 import { pure, skinnable, props, t } from '../utils';
 import ButtonLogic, { buttonState, buttonBaseState } from './ButtonLogic';
 import ButtonRenderer/*, { stringForButtonStates } */ from './ButtonRenderer';
 
-import './button.scss';
-
 // util
 const notBoth = (a, b) => !(a && b);
+const satisfyAll = (...conditions) => (props) => every(conditions.map(condition => condition(props)));
+
+// invariants
+const propsInvariants = [
+  ({ label, icon, children }) => notBoth(label || icon, children), // notBothChildrenAndLabelOrIcon
+  ({ primary, flat }) => notBoth(primary, flat), // notBothFlatAndPrimary
+  ({ fluid, circular }) => notBoth(fluid, circular), // notFluidAndCircular
+  ({ circular, icon, label }) => !circular || (icon && !label), // circularOnlyIfIconAndNotLabel
+  ({ type, primary, flat }) => notBoth(type, flat || primary) // notTypeAndItsShortucts
+];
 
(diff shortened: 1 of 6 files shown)


src/button/Button.js
```diff
@@ -72,14 +88,31 @@ export default class Button extends React.Component {
...
+    };
+
+    const wrapperStyle = {
+      display: fluid ? 'block' : 'inline-block',
+      width: fluid ? '100%' : null
+    };
+
+    const isIconButton = () => this.props.icon && !this.props.label;
+
+    const className = cx(
```
FrancescoCioria: I'm trying to switch to this convention for the state classNames:
```js
`is-${state}`  // is-default, is-icon-button, is-...
```

I added a small util in the last PR to make it easier: https://github.com/buildo/react-components/blob/master/src/utils/index.js#L6

src/button/button.scss
```diff
@@ -1,95 +1,168 @@
...
+
+$small-height: 30px !default;
+$small-horizontal-padding: 14px !default;
+$small-font-size: 15px !default;
+
+$tiny-height: 18px !default;
+$tiny-font-size: 12px !default;
+$tiny-horizontal-padding: 6px !default;
+
+$ready-hover-color: white !default;
```
FrancescoCioria: please move `-hover` to the end

src/button/button.scss
```diff
@@ -1,95 +1,168 @@
...
+$ready-hover-background: lightgray !default;
```
FrancescoCioria: same

src/button/button.scss
```diff
@@ -1,95 +1,168 @@
...
-
-  &.ready {
-    &:hover {
-      color: $button-ready-hover-color;
-      background: $button-ready-hover-background;
+  &.flat {
+    background: $flat-background;
+    border: $flat-border-width solid $flat-border-color;
+    color: $flat-color;
+    padding: 0 $flat-horizontal-padding !important; //!important because it should be ovverridden by small/tiny variations
```
FrancescoCioria: typo: replace `should` with `shouldn't`

src/button/Button.js
francescogior: it looks good, will use it here!

src/button/button.scss
francescogior: :

src/button/button.scss
francescogior: :o 👍

src/button/Button.js
francescogior: as for the util,
- I add the spread operator in the arguments so it can be used without passing an array, but use the `[].concat(...args)` trick to make it work also with an array
- I replaced the reduce with a simpler map

`export const stateClassUtil = (...classes) => cx([].concat(...classes).map(cl => `is-${cl}`));`

FrancescoCioria:
🎱 💵 🎯

PR #400, opened 2016-04-20
update versions of tcomb and revenge

giogonzo:
this looks fine from LOL point of view, testing here: https://github.omnilab.our.buildo.io/buildo/aliniq/pull/2437

@FrancescoCioria do you see this as problematic for other projects?

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,9 +38,9 @@
     "mobile-detect": "^1.2.1",
     "react-input-children": "^1.0.0",
     "react-select": "^0.6.1",
-    "revenge": "^0.3.4",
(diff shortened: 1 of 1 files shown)


FrancescoCioria:
> do you see this as problematic for other projects?

- oxway: no
- lexdoit: I don't think so (it is using `revenge` and `tcomb-react` internally but they have fixed versions in the package.json: https://github.com/buildo/lexdoit/blob/master/web/package.json#L92-L93)
- hailadoc: same as lexdoit (https://github.com/buildo/hailadoc/blob/master/web/package.json#L94-L97)

PR #403, opened 2016-04-21
#402: Refactor button (closes #402)

### LINKED ISSUE #402: Refactor button
## requirements
- refactor button in
  - a pure component
  - a wrapper that handles the internal state for onClick promises. This is what currently did ButtonLogic

- leave APIs unchanged
- provide an example where one can easily try out all the props

francescogior:
Issue #402


src/button/index.js
```diff
@@ -1 +1,2 @@
-export default from './Button';
+export default from './StatefulButton';
+export PureButton from './Button';
```
FrancescoCioria: if `StatefulButton` is the default maybe we should rename `Button` to `PureButton` and `StatefulButton` to `Button`:
I find confusing that `import Button from '../button';` will import `StatefulButton` instead of `Button`

francescogior: sure! feel free to edit!

PR #416, opened 2016-05-10
DropdownMenu menu should have an auto overflow.

ascariandrea:
That's why scrollbars appear on **AlinIQ** in `DropdownMenu`.


FrancescoCioria:
👍

PR #426, opened 2016-05-19
#425: [button] it shouldn't have hover effect nor cursor pointer in processing state (closes #425)

### LINKED ISSUE #425: [button] it shouldn't have hover effect nor cursor pointer in processing state
## description
- since it's not clickable when in processing state, it should not have hover effect nor cursor pointer

francescogior:
Issue #425


FrancescoCioria:
🌊

PR #429, opened 2016-05-20
#423: [Dropdown] multi value support is broken (closes #423)

### LINKED ISSUE #423: Dropdown: multi value support is broken
## description
The `value` passed to `react-select` [is computed](https://github.com/buildo/react-components/blob/master/src/dropdown/Dropdown.js#L86) with `valueToOption` function.

First of all, why?

Anyway, this breaks the `multi` selection feature. `valueToOption` looks for a matching value in the options, but - in the case of `multi` - `value` contains a list of comma-separated values.

## how to reproduce

i.e. perform a multiple selection on a controlled `Dropdown`.

## specs
I would just pass down `value` without calling `valueToOption`, but I'm sure there's a reason for that.
Anyone? @FrancescoCioria @francescogior @giogonzo?

Otherwise, we can just check for `multi` and pass the raw value only in that case.

FrancescoCioria:
Issue #423

## Test Plan

> These results can be screenshots, gifs, query results, or even just pointers to unit tests that successfully passed.

> examples:
  - a screenshot with your new feature active
  - if data is optional, a screenshot for when **data is missing**
  - a screenshot of the same functionality in **LOL Delphi**
  - if multiple steps are involved, a gif of **all steps**

- 900 width x 600 height and above => degraded but fully functional experience (typically, lots of scrollbars and text overflows)

> Examples:
- too many input values to test manually
- LOL Delphi not working on anybody's computer
- ? if you have other valid scenarios, please tell us :)

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -10,18 +10,37 @@
 .Select {
   position: relative;
 }
-.Select-control {
-  position: relative;
-  overflow: hidden;
-  background-color: #ffffff;
-  border: 1px solid #cccccc;
-  border-color: #d9d9d9 #cccccc #b3b3b3;
-  border-radius: 4px;
+.Select,
+.Select div,
+.Select input,
+.Select span {
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
   box-sizing: border-box;
-  color: #333333;
+}
+.Select.is-disabled > .Select-control {
+  background-color: #f9f9f9;
+}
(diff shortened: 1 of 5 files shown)


src/dropdown/dropdown.scss
```diff
@@ -169,7 +225,7 @@
   position: absolute;
   top: 100%;
   width: 100%;
-  z-index: 1000;
+  z-index: 1;
```
francescogior: reset to 1000

francescogior: forget it

PR #430, opened 2016-05-20
if multi pass array of options to react-select

FrancescoCioria:


gabro:
I can confirm this works! (tested on internal project)

PR #435, opened 2016-05-23
#432: [Modal] Create a standalone Modal component (closes #432)

### LINKED ISSUE #432: [Modal] Create a standalone Modal component
## requirements
- unify several modal implementations we made in our projects
- have an isolated modal component

## specs
- use the labonline one as starting point

## misc
{optional: other useful info}

ascariandrea:
Issue #432

### tests performed
{A Test Plan is used to show what you tested to make sure your code works fine. You should write here all that you did to test, and provide some results of your testing.

These results can be screenshots, query results, or even just pointers to unit tests that successfully passed.}

#### cross browser compatibility
{If applicable, please write on which browsers you tested this. A good test plan should include a few tests on different browsers among the supported ones.}

{At times not everything can be tested, and writing what hasn't been tested is just as important as writing what has been tested.

An example of partial test is a field displaying 4 possible values. If 3 values are tested, with screenshots, and 1 is not, then it should be mentioned here.}


ascariandrea:
I can't complete the test plan cause I have some issues on running the example.
@FrancescoCioria let me know when we can talk about this :)

FrancescoCioria:
just a note for readers: this is a temporary component needed by LexDoIt
we will replace it with a standard one asap

Issue #117, opened 2015-11-26
[TextOverflow] TextOverflow is too fragile and risky


FrancescoCioria:
## requirements
It may fail, but it should never block the entire app with an infinite loop!

## specs
given a label we have only one possible state (simple label or popover). This state can't possibly change until we have a different label -> we should check the size of children only if:
- we are currently rendering simple label template (`this.state.isOverflowing === false`)

If we receive a new `label` string as prop we reset state: -> `this.setState({ isOverflowing: true })`

PR #437, opened 2016-05-25
#436: [TextOverflow] Should update also on parent resize (closes #436)

### LINKED ISSUE #436: [TextOverflow] Should update also on parent resize
## description
as for title

## how to reproduce
- put `TextOverflow` in a resizable container

## specs
- attach to the resize event and calculate the overflow on each resize

## misc
{optional: other useful info}

FrancescoCioria:
Issue #436

### tests performed
[https://gyazo.com/865b67f71e952420b4b6c9e901f8ad1a](https://gyazo.com/865b67f71e952420b4b6c9e901f8ad1a)

#### cross browser compatibility
I didn't directly tested it on IE11 but the only browser-dependent logic is in `ResizeSensor` which works on IE11

### DIFF
--- a/src/text-overflow/TextOverflow.js
+++ b/src/text-overflow/TextOverflow.js
@@ -4,6 +4,7 @@ import omit from 'lodash/omit';
 import { props, t } from '../utils';
 import { warn } from '../utils/log';
 import Popover from '../popover/Popover';
+import ResizeSensor from '../resize-sensor/ResizeSensor';
 
 /**
  * ### Text view which, if string content is too large, trims it and shows the full content on "hover" with a `Popover` (or custom component if any)
@@ -28,9 +29,17 @@ export default class TextOverflow extends React.Component {
     this.state = { isOverflowing: false };
   }
 
-  componentDidMount = () => setTimeout(this.verifyOverflow);
+  componentDidMount() {
+    this.timeout = setTimeout(this.verifyOverflow);
+  }
 
-  componentDidUpdate = () => setTimeout(this.verifyOverflow);
+  componentDidUpdate() {
+    this.timeout = setTimeout(this.verifyOverflow);
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this.timeout);
+  }
 
   componentWillReceiveProps = (nextProps) => {
     if (nextProps.label !== this.props.label) {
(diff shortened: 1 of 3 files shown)


FrancescoCioria:
@giogonzo what do you think?

pro:
- we can keep `@pure` on wrapper
- robust (I think interval is the only robust way to verify changes of width)
cons:
- will it affect performance?? (we're using `TextOverflow` quite heavily in lol)

src/text-overflow/TextOverflow.js
```diff
@@ -54,21 +64,35 @@ export default class TextOverflow extends React.Component {
...
+    const width = this.getElementWidth(ReactDOM.findDOMNode(this.refs.text));
+    if (!this.state.width) {
+      this.setState({ width });
+    } else if (width !== this.state.width) {
+      this.setState({ width }, this.reset); // trigger this.verifyOverflow
+    }
+  };
+
   verifyOverflow = () => {
     if (this.state.isOverflowing === false && typeof window !== 'undefined') {
```
FrancescoCioria: `this.state.isOverflowing === false` is the only 100% effective way I found to avoid infinite loops ( #117 )

src/text-overflow/TextOverflow.js
```diff
@@ -54,21 +64,35 @@ export default class TextOverflow extends React.Component {
...
       const text = ReactDOM.findDOMNode(this.refs.text);
       const textWithoutEllipsis = ReactDOM.findDOMNode(this.refs.textWithoutEllipsis);
 
-      if (text && textWithoutEllipsis) {
```
FrancescoCioria: this `if` shouldn't be necessary anymore ad I'm clearing `this.interval` and `this.timeout` in `componentWillUnmount`

src/text-overflow/TextOverflow.js
```diff
@@ -54,21 +64,35 @@ export default class TextOverflow extends React.Component {
...
 
-        const isOverflowing = (textWidth < textWithoutEllipsisWidth);
-        if (isOverflowing) {
-          this.setState({ isOverflowing: true }, this.logWarnings);
-        } else {
-          this.logWarnings();
-        }
+      const isOverflowing = (textWidth < textWithoutEllipsisWidth);
+      if (isOverflowing) {
+        this.setState({ isOverflowing: true, width: null }, this.logWarnings);
```
FrancescoCioria: I'm setting `width` to `null` to reset the behavior of `this.verifySize`

src/text-overflow/TextOverflow.js
```diff
@@ -25,12 +25,22 @@ export default class TextOverflow extends React.Component {
...
   constructor(props) {
     super(props);
-    this.state = { isOverflowing: false };
+    this.state = { isOverflowing: false, width: null };
   }
 
-  componentDidMount = () => setTimeout(this.verifyOverflow);
+  componentDidMount() {
+    setTimeout(this.verifyOverflow);
+    this.interval = setInterval(this.verifySize, 100);
```
FrancescoCioria: `100` looks very arbitrary.... maybe we could increase it?

src/text-overflow/TextOverflow.js
```diff
@@ -25,12 +25,22 @@ export default class TextOverflow extends React.Component {
...
   constructor(props) {
     super(props);
-    this.state = { isOverflowing: false };
+    this.state = { isOverflowing: false, width: null };
   }
 
-  componentDidMount = () => setTimeout(this.verifyOverflow);
+  componentDidMount() {
+    this.timeout = setTimeout(this.verifyOverflow);
+    this.interval = setInterval(this.verifySize, 100);
```
FrancescoCioria: (maybe make `100` overridable through `props`?)

FrancescoCioria:
@giogonzo tested successfully on LOL (table header resize)

PR #440, opened 2016-05-30
#439: Badge label should be react children (closes #439)

### LINKED ISSUE #439: Badge label should be react children
because of https://github.omnilab.our.buildo.io/buildo/aliniq/issues/2787 (badge should show mixed font-weight text in the label) we need the `Badge` component to accept `ReactChildren` as label

francescogior:
Issue #439

## Test Plan
- it should accept a ReactChildren as label


FrancescoCioria:
👶

PR #448, opened 2016-05-31
#447: [ResizeSensor] throws error on detach (closes #447)

### LINKED ISSUE #447: [ResizeSensor] throw error on detach
## description
throws error when updating after children have changed due to the detach failing

## how to reproduce

## specs
after [inspecting the lib code](https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js) I can say that:
- the cleaning `detach` is useless as everything is stored inside the `element` and there's no `removeEventListener` in the code
  - when attaching it creates variables in `element` and add a children node inside it
  - when `children` changes the previous `element` is removed from the DOM --> we can skip the "cleaning"
- the attach action works only if `!element.resizedAttached`
  - we should make a redundant check just in case before attaching a new resize sensor

## misc
{optional: other useful info}

FrancescoCioria:
Issue #447

### tests performed
- button + textoverflow example now works without errors
- tested on LOL too

I assumed a couple of things, read them in issue specs

### DIFF
--- a/src/resize-sensor/ResizeSensor.js
+++ b/src/resize-sensor/ResizeSensor.js
@@ -12,35 +12,19 @@ import _ResizeSensor from 'css-element-queries/src/ResizeSensor';
 export default class ResizeSensor extends React.Component {
 
   attachResizeSensor = () => {
-    this.resizeSensor = new _ResizeSensor(
-      ReactDOM.findDOMNode(this),
-      this.onResize
-    );
-  }
-
-  detachResizeSensor = () => {
-    this.resizeSensor.detach(ReactDOM.findDOMNode(this));
-  }
-
-  updateResizeSensor = () => {
-    if (this.resizeSensor) {
-      this.detachResizeSensor();
+    const element = ReactDOM.findDOMNode(this);
+    if (!element.resizedAttached) {
+      this.resizeSensor = new _ResizeSensor(element, this.onResize);
     }
-    this.attachResizeSensor();
   }
 
   componentDidMount() {
     _ElementQueries.listen();
-    this.updateResizeSensor();
+    this.attachResizeSensor();
   }
 
   componentDidUpdate() {
-    this.updateResizeSensor();
-  }
-
-  componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


src/resize-sensor/ResizeSensor.js
```diff
@@ -1,51 +1,42 @@
...
     _ElementQueries.listen();
-    this.updateResizeSensor();
+    this.attachResizeSensor();
   }
 
   componentDidUpdate() {
-    this.updateResizeSensor();
-  }
-
-  componentWillUnmount() {
```
giogonzo: I think your assumptions on not needing to `detach()` are correct, but what's the cost of adding a call to detach here? So that we are 100% sure we're cleaning everything up...

`if !!findDOMNode()` should be enough to make ti safe?

giogonzo:
👍 except for inline proposal

src/resize-sensor/ResizeSensor.js
FrancescoCioria: In `componentDidUpdate` this is not true as `findDOMNode` will always be `true` because of  the newly `this.refs.text` already being rendered... (we should check if it contains the resize element added by `ResizeSensor` to understand if there's something to detach or not).

Nevertheless we *should* be able to call it on `componentWillUnmount`, do you think it could be enough adding it in `cwum`?

giogonzo: yes didn't specify but I'm talking about `cwum`

FrancescoCioria: nope, `Button` + `TextOverflow` throws an error for the reason I wrote above... If you agree I would leave it as it is and, if we have performance issues, find a way to determine if contains or not the resize sensor element (I wanted to avoid this as it would need a css query selector which we'll then need to keep up to date)

FrancescoCioria: opened https://github.com/buildo/react-components/issues/453 : we can continue the discussion there (so to unblock other PRs and LOL which is currently broken)

PR #457, opened 2016-06-02
#456: [ResizeSensor] calling this.props.onResize when already unmounted (closes #456)

### LINKED ISSUE #456: [ResizeSensor] calling this.props.onResize when already unmounted
## description
`ResizeSensor` may call `this.props.onResize` when already unmounted. this could lead to errors

## how to reproduce

## specs
add some check to avoid calling `this.props.onResize` if unmounted

## misc
{optional: other useful info}

FrancescoCioria:
Issue #456

### tests performed
- works like before
- fixed warning in LOL


giogonzo:
:+1:

PR #458, opened 2016-06-02
#455: [FocusableView] optionally debounce focus/blur events (closes #455)

### LINKED ISSUE #455: FocusableView: optionally debounce focus/blur events
## requirements
In some circustances a `FocusableView` may quickly change focus 2/3 times, this causes a waterfall re-render. To avoid/mitigate the re-render waterfall we should optionally debounce blur/focus events and keep the last one

## specs
- add optional prop debounce millis
  - if it exists debounce focus/blur events with `_.debounce` (is it possible to keep only the last one?)

## misc
{optional: other useful info}

FrancescoCioria:
Issue #455

### tests performed
`debounce: undefined` --> no debounce
`debounce: 1000` --> focus/blur events are merged and debounced of 1000ms
- rapidly focus/blur/focus/blur --> onFocus/onBlur never called --> no rerender even without `@pure`

### DIFF
--- a/src/focusable-view/FocusableView.js
+++ b/src/focusable-view/FocusableView.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import cx from 'classnames';
+import debounce from 'lodash/debounce';
 import { props, t, skinnable } from '../utils';
 
 export const Props = {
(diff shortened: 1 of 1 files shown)


FrancescoCioria:
@giogonzo ping

PR #460, opened 2016-06-06
#459: Disabled state for confirmation input and dropdown (closes #459)

### LINKED ISSUE #459: Disabled state for confirmation input and dropdown
- add strict: false to conf input props of conf input
- explicitly disabled prop of conf input
- add style for disabled state of conf input
- add sass var for disabled opacity of conf input and dropdown

francescogior:
Issue #459


FrancescoCioria:
🐼

PR #461, opened 2016-06-11
fix stateless toggle

giogonzo:
`value` and `onChange` props were not considered

## test plan

didn't test the valueLink version yet


giogonzo:
discussed offline 👍

PR #467, opened 2016-06-25
#466: use ResizeSensor for Table autosize (closes #466)

### LINKED ISSUE #466: use ResizeSensor for Table autosize
## requirements
`Table` autosize interval is a performance killer, we should use `ResizeSensor` instead

## specs
- add prop `debounce` to `ResizeSensor` to easily support debouncing
- wrap `Table` with `ResizeSensor` and remove autosize interval

## misc

FrancescoCioria:
Issue #466

### tests performed
[https://gyazo.com/bee37aee7d86e83ed76550653ac497c4](https://gyazo.com/bee37aee7d86e83ed76550653ac497c4)

**tested on QIA too**

### DIFF
--- a/src/resize-sensor/ResizeSensor.js
+++ b/src/resize-sensor/ResizeSensor.js
@@ -51,11 +51,11 @@ export default class ResizeSensor extends React.Component {
     return this.props.children;
   }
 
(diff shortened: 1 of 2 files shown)


francescogior:
nice!

PR #469, opened 2016-06-26
#468: make TextOverflow optionally lazy (closes #468)

### LINKED ISSUE #468: make TextOverflow optionally lazy
## requirements
`TextOverflow.verifyOverflow` is slow, especially on IE11 where DOM size functions (`offsetWidth`, `getComputedStyle`...) are dozens of time slower than Chrome.
In some situations (ex: tables!) we may prefer to use a lazy approach and *pay* the `verifyOverflow` function only when we need it

## specs
- add boolean prop `lazy`
- if `lazy === true`
  - disable `verifyOverflow` in `cdm` and `cdu`
  - add `onMouseEnter` callback to `text` node
    - call `verifyOverflow`
    - set `state.isHovering` to `true`
  - add `onMouseLeave` callback to `text` node
    - set `state.isHovering` to `false`
  - pass `popover.isOpen: this.state.isHovering` to `Popover`

## misc
{optional: other useful info}

FrancescoCioria:
Issue #468

### tests performed
- tested every example with both `lazy: true` and `lazy: false` -> everything worked as expected
- tested on QIA -> ✅ 
- tested on AMS -> ✅ 

### DIFF
--- a/examples/components/index.js
+++ b/examples/components/index.js
@@ -1,3 +1,4 @@
+import { sortBy } from 'lodash';
 import examplesJSON from 'raw!./examples.json';
 import './styles';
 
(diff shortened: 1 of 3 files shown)


francescogior:
💇 🐹 🔨 🍔

PR #471, opened 2016-06-28
#470: Add optional delay to TextOverflow (closes #470)

### LINKED ISSUE #470: Add optional delay to TextOverflow
## requirements
add optional integer prop `delayWhenLazy` to delay the tooltip in `TextOverflow` when in "lazy" mode.
this is useful to avoid defects as fast mouse hovering may cause the browser to "forget" about the `onMouseLeave` event as it's doing heavy calculations (when lazy the overflow calculation is done in `onMouseEnter`!)

## specs
- add prop `delayWhenLazy`
  - defaults to `100`
  - if it exists debounce `onMouseEnter` events of that amount
  - `onMouseLeave` should not be delayed
  - `onMouseLeave` should invalidate previous `onMouseEnter` events not yet called

## misc
{optional: other useful info}

FrancescoCioria:
Issue #470

### tests performed
[https://gyazo.com/23513d71544a9aa9ed51be6945c3e605](https://gyazo.com/23513d71544a9aa9ed51be6945c3e605)

PS: tested on QIA too --> the defect (tooltip remaining open) has gone 🎉 

### DIFF
--- a/examples/components/text-overflow/Default.example
+++ b/examples/components/text-overflow/Default.example
@@ -18,6 +18,7 @@ const Example = React.createClass({
           label={this.state.text}
           popover={{ position: 'bottom', anchor: 'left' }}
           lazy
+          delayWhenLazy={1000}
         />
       </button>
     );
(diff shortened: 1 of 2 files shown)


francescogior:
🗼 !

PR #474, opened 2016-06-29
#472: [Meter] Range background color (closes #472)

### LINKED ISSUE #472: [Meter] Range background color
## requirements
It should be possible to specify meter bar background color based on different values

## specs
- a `range` should accept a new prop named `backgroundColor`
- meter should set the bar background color based on the value specified in the range within which the value falls
- `meter` should accept a new prop `baseBackgroundColor`, for fallback background color

## misc
{optional: other useful info}

veej:
Issue #472

### tests performed
- `npm run test` ✅ 

screen shot 2016-06-29 at 12 51 51 pm

### DIFF
--- a/examples/components/meter/Meter.example
+++ b/examples/components/meter/Meter.example
@@ -1,7 +1,7 @@
 class Example extends React.Component {
   render() {
     const ranges = [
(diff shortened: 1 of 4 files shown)


FrancescoCioria:
Great work!

PR #487, opened 2016-08-01
#486: Button should be clickable only when is ready (closes #486)

### LINKED ISSUE #486: Button should be clickable only when is ready
It was mostly hidden because the `StatefulWrapper` wrapper fixed this

francescogior:
Issue #486

## Test Plan


giogonzo:
👍

PR #489, opened 2016-08-02
#488: [Button] children must be string, and code style (closes #488)

### LINKED ISSUE #488: BREAKING: [Button] children must be string, and code style
Because of this code,
https://github.com/buildo/react-components/blob/master/src/button/Button.js#L126-L129
we must force `children` to be a `String` in `@prop`. The component would break if a `ReactChildren` is passed as children

francescogior:
Issue #488

## Test Plan

tested this branch on qia

### DIFF
--- a/src/button/Button.js
+++ b/src/button/Button.js
@@ -112,6 +112,8 @@ const defaultIcons = {
   error: 'exclamation'
 };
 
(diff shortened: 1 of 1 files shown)


giogonzo:
💯

PR #499, opened 2016-08-08
#498: subtle change in ui behaviour of table scrollbars (closes #498)

### LINKED ISSUE #498: subtle change in ui behaviour of table scrollbars
update scrollbar "thumb" style to make it consistent *while* dragging

in other words, copy all of thumb `:hover` style also to `:active`

francescogior:
Issue #498


giogonzo:
🍎

PR #502, opened 2016-08-12
#494: [KitchenSink] Readme generator (closes #494)

### LINKED ISSUE #494: [KitchenSink] Readme generator
← #490

## requirements
- generate component `Readme.md` by extracting documentation from component:
  - `props` definition
  -  `description` (defined as comment maybe)

## specs
- try `parse` from `tcomb-react` utility method
- implement custom `toMarkdown` function (see old one in `generateReadme.js` for help)

## misc
- https://github.com/gcanti/tcomb-react#the-tomarkdown-function

ascariandrea:
Issue #494

I created a new version of `generateReadme.js` file following the approach given by `tcomb-react` to extract component documentation.
The script is testable by:
```shell
# to generate README.md file for each component
$ npm run generate-all-readmes 
```

The npm task doesn't work as it is cause the components don't have the correct documentation structure to be parsed, cause I've defined strict rules in order to be sure every component is documented rightly, and the rules are:
- component must have a name (now it's extracted from the the class name)
- component must have a description (even a simple one, but it's required from my pov)
- each component props must be documented

You can add comments 
```js
/**
 * Component description here
 * @param prop1 - this is the prop 1 description 
 * @param prop2 - this is the prop 2 description
 * ...
 */
```
to a component and test with 
```shell
# to generate README(s).md for a specific component
$ babel-node scripts/generateReadme.js src/${componentfolder} 
```

**N.B.** the file is full of log for debug purpose.

### DIFF
--- a/scripts/generateReadmes.js
+++ b/scripts/generateReadmes.js
@@ -61,6 +61,41 @@ const walkThroughDirs = (dir, parent, cb) => {
   }
 };
 
+const walkThroughFirstLevelDirs = (dir, cb) => {
+  if (fs.statSync(dir).isDirectory()) {
+    log(`Walking through "${dir}"`);
+
+    const results = fs.readdirSync(dir);
+
+    log(`"${dir}" contains: ${results.join('\n')}`);
+
+    results.forEach(r => {
+      const pathToCheck = path.join(dir, r);
+      log(`Checking path: ${pathToCheck}`);
+      if (fs.statSync(pathToCheck).isDirectory()) {
+        const mainFileName = `${upperFirst(camelCase(r))}.js`;
+
+        const mainFilePath = path.join(dir, r, mainFileName);
+
+        try {
+          if (fs.statSync(mainFilePath).isFile()) {
+            log(`"${r}" contains a main file: "${mainFileName}"!`);
+            cb(pathToCheck, mainFileName);
+          } else {
+            log(`"${r}" does not contain a main file...`);
+          }
+        } catch (e) {
+          log(e);
+        }
+      } else {
+        log(`"${r}" is a file: ignoring!`);
+      }
+    });
+  } else {
+    log(`"${dir}" is not a directory!`);
+  }
+};
+
 const root = process.argv[2] || 'src';
 
-walkThroughDirs(path.resolve(root), '', generateReadMe);
+walkThroughFirstLevelDirs(path.resolve(root), generateReadMe);
--- a/scripts/generateReadmes.js
+++ b/scripts/generateReadmes.js
@@ -4,6 +4,8 @@ import t from 'tcomb';
 import parse from './parse';
 import template from './template';
 import debug from 'debug';
+import upperFirst from 'lodash/upperFirst';
+import camelCase from 'lodash/camelCase';
 
 debug.enable('brc:*');
 const log = debug('brc:generate-readmes');
@@ -17,11 +19,6 @@ const checkDocs = (json, fileName) => {
     throw new Error(`Missing description for file ${fileName}.`);
   }
 
-};
-
-const kebabToPascalCase = value => {
-  const words = value.split('-').map(w => w.charAt(0).toUpperCase() + w.slice(1).toLowerCase());
-  return words.join('');
 };
 
 const generateReadMe = (dir, fileName) => {
@@ -50,7 +47,7 @@ const walkThroughDirs = (dir, parent, cb) => {
       .forEach(r => {
         const pathToCheck = path.join(dir, r);
         log(`Checking path: ${pathToCheck}`);
-        if (fs.statSync(pathToCheck).isFile() && r === `${kebabToPascalCase(parent)}.js`) {
+        if (fs.statSync(pathToCheck).isFile() && r === `${upperFirst(camelCase(parent))}.js`) {
           log(`${r} is a file!`);
           cb(dir, r);
         } else if (fs.statSync(pathToCheck).isDirectory()) {
--- a/scripts/generateReadmes.js
+++ b/scripts/generateReadmes.js
@@ -19,6 +19,11 @@ const checkDocs = (json, fileName) => {
 
 };
 
+const kebabToPascalCase = value => {
+  const words = value.split('-').map(w => w.charAt(0).toUpperCase() + w.slice(1).toLowerCase());
+  return words.join('');
+};
+
 const generateReadMe = (dir, fileName) => {
   const filePath = path.join(dir, fileName);
   log(`Extracting docs from ${filePath}`);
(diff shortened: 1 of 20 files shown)


FrancescoCioria:
> each component props must be documented

I agree 99% with this: I would give a *default* documentation to general props `className`, `style` and `id` (the default documentation should of course be overridable) as in most components they have the same exact purpose (pass class, id, inline-style to first html tag).

scripts/parse.js
```diff
@@ -0,0 +1,103 @@
...
+// (component: ReactComponent) => t.String
+function getComponentName(component) {
+  return component.name;
+}
+
+// (component: ReactComponent) => TcombType
+function getPropsType(component) {
+  const propTypes = component.propTypes;
+  const props = {};
+  Object.keys(propTypes).forEach((k) => {
```
FrancescoCioria: this should be a `.reduce` IMO

scripts/parse.js
```diff
@@ -0,0 +1,103 @@
...
+      ret.props[tag.name] = tag.description || null;
+    }
+  });
+  return ret;
+}
+
+// (exports: t.Object) => ReactComponent
+function getComponent(defaultExport) {
+  const def = defaultExport['default'] || defaultExport; // eslint-disable-line dot-notation
+  // if (defaultExport['default']) { // eslint-disable-line dot-notation
```
FrancescoCioria: old code forgotten?

scripts/parse.js
```diff
@@ -0,0 +1,103 @@
...
+    const comments = getComments(path);
+    const descriptions = getDescriptions(comments);
+    const type = toObject(getPropsType(component));
+    const props = type.kind === 'refinement' ? type.type.props : type.props;
+    const defaultProps = getDefaultProps(component);
+    const name = getComponentName(component);
+    const propsSchema = Object.keys(props).reduce((acc, prop) => {
+      const comment = find(comments.tags, (t) => t.name === prop);
+      if (!comment) {
+        throw new Error(`Missing description for prop '${prop}' in '${name}'.`);
```
FrancescoCioria: if you agree I would add here a default fallback for `className`, `id` and `style`. Do you see any problem with it?

scripts/parse.js
```diff
@@ -0,0 +1,103 @@
+
```
FrancescoCioria: empty line :)

scripts/generateReadmes.js
```diff
@@ -0,0 +1,65 @@
...
+  if (fs.statSync(dir).isDirectory()) {
+    log(`Walking through ${dir}`);
+    fs.readdir(dir, (err, results = []) => {
+      log(`${dir} contains: ${results.join('\n')}`);
+      results
+      .filter(r => r !== 'index.js' && ['', '.js'].indexOf(path.extname(r)) > -1)
+      .forEach(r => {
+        const pathToCheck = path.join(dir, r);
+        log(`Checking path: ${pathToCheck}`);
+        if (fs.statSync(pathToCheck).isFile() && path.extname(r) === '.js') {
```
FrancescoCioria: I have some concerns about this logic.
For which files do we want to create a README.md? I thought only the main component file.
How do you handle folders with multiple components in it? last component wins?
also, if we have a folder like:
```
- /component
  - MainComponent.js
  - /sub-components
    - Header.js
    - Body.js
    - Footer.js
```
are you generating a README.md also for the folder `sub-components`? I think we shouldn't.

If you want developers to document also minor components we can think of simple CI validation check which controls if every component is documented properly (you already implemented the logic for this) without creating README.md files.

what do you think @ascariandrea?

veej: We could use a JSON file with the list of main components, or use a particular comment to start the description of only the components we want to include in the process... any better idea?

ascariandrea: I think we need to separate the **sub-component** concept in:
- **sub-component** created to reduce the parent component code by splitting it in sub-files, not usable as stand-alone 
- **sub-component** intended as **reusable component**

In the first case I agree no `README` file should be generated for the **sub-components**, but for the second case I think a proper `README` is needed.

e.g. `Panel` component:
- Used in `TabbedPanel` (**reusable component**)
- Usable as it is (exported)

In this case I think each component need a proper folder and its `README`.

If I'm not wrong, we don't have components with **sub-components** not exported, but I would manage the case with a strong folder naming convention and leave the scripts as much as dynamic as possible.

FrancescoCioria: > and leave the scripts as much as dynamic as possible.
I strongly disagree with this 😛 

I'd rather have a simple but reliable script which work in fewer cases

veej: Actually we have few `.js` files that are not even components, and at the moment they're processed by the script (e.g. `scroll/easingFunctions.js`), so IMO a method to filter files that need a `README` is mandatory

ascariandrea: @FrancescoCioria What's your suggestion about that?

FrancescoCioria: I suggest to start with a script that creates a README only for the main js file of a folder.
The "main" file is the one in the folder with the pascal-case version of the folder name. if there is no file with this name we could simply ignore the folder and make a log (ex: `/utils` will have no file named `Utils.js`)

scripts/generateReadmes.js
```diff
@@ -19,6 +19,11 @@ const checkDocs = (json, fileName) => {
 
 };
 
+const kebabToPascalCase = value => {
```
FrancescoCioria: (you could easily replace this with `_.upperFirst(_.camelCase(value))`)

scripts/generateReadmes.js
ascariandrea: I agree with this approach, my point was to not use a manually edited list of components who need READMEs.
But I'm strongly convinced if a component is exported to be usable it needs a README too, so components like `TabbedPanel` need to have their own folder.

veej: I modified the script to look only for the main `.js` file in a dir, as @FrancescoCioria described. I kept the recursion over sub-directories, so it should be easy to move sub/parallel components in their own sub-dir in order to be processed by the script. 
I think we should decide whether it's worth presenting sub-components in a dedicated page or in a sub-section of the main component.

ascariandrea: I'd put sub-components in a sub-section of main component to give users the context they need.

FrancescoCioria:
UPDATE

1. script generates readme only for main file in first-level folders (we'll discuss sub-components in new issue)
2. enums use a custom formatter instead of the name
3. descriptions **can't** contain "|" (it would break the table template)
4. fixed KitchenSink components

PR #504, opened 2016-08-16
#503: Remove PanelMenu component (closes #503)

### LINKED ISSUE #503: Remove PanelMenu component
## requirements
`PanelMenu` component is unused and broken at this time (since it tries to import and use a non-existent `TooltippedIcon` component). We could safely remove it (it's not even exported in `Panel/index.js`)

## specs
- we could safely remove `PanelMenu.js` and any other reference to it (exports, style imports, ...)

## misc
{optional: other useful info}

veej:
Issue #503

## Test Plan


FrancescoCioria:
💀  👍

PR #506, opened 2016-08-17
#505: Remove styles import (closes #505)

### LINKED ISSUE #505: Remove styles import
## requirements
Importing `.scss` files should be delegated to the final user of the react-component, and not to the component itself. Moreover, component `.scss` file should define variables that can be overwritten when the style is imported. 

## specs
- Remove `import *.scss` from every component `.js` file
- Define component-specific variables in SASS files. `.scss` files should not import from `variables.scss` 

## misc
{optional: other useful info}

veej:
Issue #505

### tests performed
- `npm start`
- `examples look fine` ✅ 

### DIFF
--- a/src/Divider/divider.scss
+++ b/src/Divider/divider.scss
@@ -1,8 +1,8 @@
-@import '../utils/theme/variables.scss';
+$color: #E6E6E6;  // pearl
 
 .divider {
-  background-color: $pearl;
-  color: $pearl;
+  background-color: $color;
+  color: $color;
   &.small {
     margin: 5px;
   }
--- a/src/Divider/divider.scss
+++ b/src/Divider/divider.scss
@@ -1,4 +1,4 @@
-$color: #E6E6E6;  // pearl
+$color: #E6E6E6 !default;  // pearl
 
 .divider {
   background-color: $color;
--- a/src/DropdownMenu/menu.scss
+++ b/src/DropdownMenu/menu.scss
@@ -1,4 +1,9 @@
-@import '../utils/theme/variables.scss';
+$background: white;
+$box-shadow-color: #628292;             // $lynch
+$menu-title-color: #bdc2ca;             // $silver
+$menu-item-title-color: #2c3a40;        // $charcoal
+$menu-item-title-color-hover: #628292;  // $lynch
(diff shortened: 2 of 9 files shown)


src/Divider/divider.scss
```diff
@@ -1,8 +1,8 @@
-@import '../utils/theme/variables.scss';
+$color: #E6E6E6;  // pearl
```
giogonzo: shouldn't we set all of these variables with a `!default` flag to make them overridable?

e.g.
```scss
$color: #E6E6E6 !default;
```

src/Panel/tabbedPanel.scss
```diff
@@ -1,4 +1,6 @@
-@import '../utils/theme/variables.scss';
+$tabbed-panel-tab-color: #3f4e49 !default;       // $gunmetal
```
FrancescoCioria: we're already in the context of "tabbed-panel", you can simply rename them as "tab-color" and "tab-color-active"

src/DropdownMenu/menu.scss
```diff
@@ -1,14 +1,19 @@
-@import '../utils/theme/variables.scss';
+$background: white !default;
+$box-shadow-color: #628292 !default;             // $lynch
+$menu-title-color: #bdc2ca !default;             // $silver
```
FrancescoCioria: also here, you could rename "menu-title-color" in "title-color".

(I'm not sure if "menu-item-..." should be renamed to "item-..." I'll leave the decision to you)

src/Panel/tabbedPanel.scss
giogonzo: is this true even in terms of `.scss` (or better, `sass-loader`) semantics? In other words, are we sure sass-loader is compiling each top-level `.scss` import alone or we must take care of namespacing ourselves?

giogonzo: couldn't find an answer to this but I assume sass-loader is spawning a node-sass compilation for each top level import of a `.scss` from js, so we should be ok

FrancescoCioria:
@veej your test plan is empty, did you run the examples to see that you're correctly importing every sass?

FrancescoCioria:
nevermind, I tested them and it looks ok

PR #514, opened 2016-08-22
#513: dropdown's menu's scrollbars hide menu when clicked in ie11 (closes #513)

### LINKED ISSUE #513: dropdown's menu's scrollbars hide menu when clicked in ie11
## specs
- solved by https://github.com/JedWatson/react-select/pull/539/files, but the patch isn't included in the `0.7` we are using here
- thus the defect will be solved by the migration to `1.0`. In the meantime, override the `handleInputBlur` will do the job

francescogior:
Issue #513


FrancescoCioria:
👍 👍 👍

PR #516, opened 2016-08-24
#515: Add gemini ScrollView (closes #515)

### LINKED ISSUE #515: Add gemini ScrollView
## requirements
in AlinIQ we created a component `ScrollView` which uses gemini scrollbar to standardize the scrollbars through different browsers.

As we're already starting to use it in other projects (ex: IperCron) we should move the code here to have a standard Gemini ScrollView component.

## specs
- copy-paste component from AlinIQ
  - add sass-variables API
- add example

## misc
This component name will conflict with `scroll/ScrollView`.
The two components have different independent features --> they can (and will) be merged in a single component soon.

For now it's ok to export this new one from `src/index.js` as it's the one most projects will use in the short term

francescogior:
Issue #515

## Test Plan

- added ScrollView (as it was, in other project)
- sass exposes style variables
  - $scrollbarSize: 7px !default;
  - $scrollbarSizeBig: 9px !default;
  - $scrollbarColor: grey !default;
  - $scrollbarContainerColor: white !default;

- added example

### DIFF
--- a/examples/components/ScrollView/ScrollView.example
+++ b/examples/components/ScrollView/ScrollView.example
@@ -1,0 +1,15 @@
+class Example extends React.Component {
+
+  render() {
+    return (
+      <div style={{ height: 300 }}>
+        <ScrollView>
+          <div style={{ height: 600}}>
+            <img alt='scroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scroll' style={{ width: '100%'}} src='https://media.giphy.com/media/26BRERwHtgJTf7rTG/giphy.gif' />
+          </div>
+        </ScrollView>
+      </div>
+    );
+  }
+
+}
--- a/examples/components/ScrollView/ScrollView.example
+++ b/examples/components/ScrollView/ScrollView.example
@@ -5,7 +5,7 @@ class Example extends React.Component {
       <div style={{ height: 300 }}>
         <ScrollView>
           <div style={{ height: 600}}>
-            <img alt='scroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scroll' style={{ width: '100%'}} src='https://media.giphy.com/media/26BRERwHtgJTf7rTG/giphy.gif' />
+            <img alt='scrollscroll' style={{ width: '100%'}} src='https://media.giphy.com/media/26BRERwHtgJTf7rTG/giphy.gif' />
           </div>
         </ScrollView>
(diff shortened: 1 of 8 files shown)


examples/components/ScrollView/ScrollView.example
```diff
@@ -0,0 +1,15 @@
+class Example extends React.Component {
+
+  render() {
+    return (
+      <div style={{ height: 300 }}>
+        <ScrollView>
+          <div style={{ height: 600}}>
+            <img alt='scroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scrollscroll scroll' style={{ width: '100%'}} src='https://media.giphy.com/media/26BRERwHtgJTf7rTG/giphy.gif' />
```
FrancescoCioria: please don't spam in the live playground :P (can you remove those scroll scrollscroll scroll scrollscroll?)

src/ScrollView/scrollView.scss
```diff
@@ -0,0 +1,113 @@
+@import '~gemini-scrollbar/gemini-scrollbar.css';
+@import '~sass-flex-mixins/src/sass-flex-mixins.scss';
+
+$scrollbarSize: 7px !default;
```
FrancescoCioria: sass variables should use kebab-case

src/ScrollView/scrollView.scss
```diff
@@ -0,0 +1,113 @@
...
+$scrollbarSizeBig: 9px !default;
```
FrancescoCioria: where is this variable used? the name is not very clear... (I thought we had two available themes: standard/big)

examples/components/ScrollView/ScrollView.example
francescogior: i put a picture (remote url) in the scrollable area and this would be the alternative text if there's no internet. The picture can be considered spammy too

Do you like it?

src/ScrollView/scrollView.scss
francescogior: will fix asap

src/ScrollView/scrollView.scss
francescogior: down below, right? It's because when in `:hover` and `:active` state, the scrollbars is thicker.

FrancescoCioria: do you think we need a variable? couldn't we dynamically define as `$scrollbar-size + 123px`?

francescogior: I think we don't need but it should be more like `$scrollbar-size * 1.20`, that is valid sass but would leave decimals

francescogior:
@FrancescoCioria pushed a commit to fix what you commented

FrancescoCioria:
👍 🌷

PR #536, opened 2016-08-31
#535: Avoid SASS variables collisions when override default style (closes #535)

### LINKED ISSUE #535: Avoid SASS variables collisions when override default style
## description
As a standard way to override the style of a component, we follow this procedure in component's custom .scss file: 

1. import project's `variables.scss`
2. override component's SASS variables
3. import component's .scss file
4. write custom CSS rules

As a consequence of step 3, we're importing BRC's `variables.scss` too, overriding project's variables with colliding names.

## how to reproduce
- Define a `variables.scss` in your project using a variable name already used by BRC's `variables.scss` (like $coolGrey), but with a different value.
- redefine a component's style following the usual procedure. In step 4, write a custom rule that use the colliding variable ($coolGrey)
- Your custom style will use the definition of $coolGrey from BRC, instead of the one you defined for your project.

## specs
Rename every BRC SASS variable defined in `variables.scss` prepending a `brc-` to the variable name. This will make collisions less likely.

## misc
{optional: other useful info}

veej:
Issue #535

### tests performed
Everything in Showroom should work as before

### DIFF
--- a/src/Divider/divider.scss
+++ b/src/Divider/divider.scss
@@ -1,6 +1,6 @@
 @import '../utils/theme/variables.scss';
 
-$color: $silver !default;
+$color: $brc-silver !default;
 
 .divider {
   background-color: $color;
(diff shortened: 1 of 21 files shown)


FrancescoCioria:
🍰

PR #537, opened 2016-08-31
#517: Update react-select to v1.0.0-rc1 (closes #517)

### LINKED ISSUE #517: Update react-select to v1.0.0-rc1
## requirements
`react-select@1.0.0-rc1` brings many new features we'd like to have in our `Dropdown` component (ex: #520, #513, autoBlur...) -> we should upgrade to v1.0.0-rc1!

## specs
- remove these hacks:
  - TODO
- fix CSS overrides (layout/CSS has changed in v1.0.0-rc1)

## misc
{optional: other useful info}

francescogior:
Issue #517

### DIFF
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -6,6 +6,8 @@ import omit from 'lodash/omit';
 import cx from 'classnames';
 import { warn } from '../utils/log';
 
+const isEmptyArray = x => t.Array.is(x) && x.length === 0;
+
 export const Props = {
   value: t.maybe(t.union([t.Number, t.String, t.Object, t.list(t.Object)])),
   valueLink: t.maybe(t.struct({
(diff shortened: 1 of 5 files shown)


FrancescoCioria:
@francescogior I added three commits, can you take a look at them?

francescogior:
@FrancescoCioria added two commits

src/dropdown/Dropdown.js
```diff
@@ -102,9 +102,18 @@ export default class Dropdown extends React.Component {
     return onChange(value);
   }
 
+  onInputKeyDown = (e) => {
```
FrancescoCioria: does it event make sense to have a dropdown with an empty list of options? wouldn't it be better to improve `@props` and forbid `props.options` to be empty?

FrancescoCioria: discussed offline: there cases where it does make sense (or simply where it may happen) to have a dropdown with no options

FrancescoCioria:
👍 👍 👍

PR #546, opened 2016-09-02
#545: Remove FlexView from b-r-c (closes #545)

### LINKED ISSUE #545: Remove FlexView from b-r-c
`FlexView` has been officially moved to its own shyn repo `react-flexview` (https://github.com/buildo/react-flexview).

Let's remove `flex/` from b-r-c to avoid confusion.

FrancescoCioria:
Issue #545

### tests performed
- `rm -rf node-modules && npm i && npm start`
  - app starts correctly
  - navigate to **every component**
    - components work correctly


giogonzo:
💯

PR #550, opened 2016-09-05
#544: [KitchenSink] Panel page and examples (closes #544)

### LINKED ISSUE #544: [KitchenSink] Panel page and examples
← #490

## requirements
Just like the `Button` page, we'd like to have a similar page for `Panel`.
Examples must include default panel, panel with header, collapsable panels, dark theme and tabbed panel.

## specs
Mockups could be seen [here](https://zpl.io/Gp1hc)

## misc
{optional: other useful info}

veej:
Issue #544


src/kitchen-sink/KitchenSink.js
```diff
@@ -35,7 +35,8 @@ export const Props = {
  * @param iso - wheter the kitchen-sink render examples in a fake isomorphic environment
  * @param header - renderable node used as header in component page
  * @param footer - renderable node used as footer in component page
- * @param loading - wheter it's loading or not
+ * @param loading - whether it's loading or not
+ * @param children - custom content
```
FrancescoCioria: I already added this on `master` last week, try a rebase

src/Panel/panel.scss
```diff
@@ -92,6 +92,14 @@ $content-title-color: $brc-darkGrey !default;
     border-left: 1px solid $border-color;
   }
 
+  &.floating {
```
FrancescoCioria: :(
Is this PR going to be "breaking" again? didn't you fixed the styles in the last one?

veej: This in particular shouldn't be a breaking change: new default style adds a border-bottom to the header, ending in a double border when collapsed. Old style didn't have border-bottom, so it should be a safe change. 

Not sure about the margin-left change few lines below (do we have some collapsable-left panel with the rotated title visible when collapsed)? 
Actually this change could be reverted and customized in the example

FrancescoCioria: > do we have some collapsable-left panel with the rotated title visible when collapsed

yes, in AMS. If it's not a problem I would revert it to avoid a breaking change. As it's a small fix we can do it later on once we've finished the re-styling

FrancescoCioria:
🎨 👍

PR #576, opened 2016-09-07
Switch over to Jest for tests

gabro:
Caveats: currently using enzyme to support snapshot testing, since `react-test-renderer` requires React 15.


FrancescoCioria:
@gabro CI is failing for a strange reason (locally `npm test` works successfully) area you sure that `jest` is available globally?

gabro:
it's just eslint not knowing about the global `jest` object. Fixed in the latest commit

PR #579, opened 2016-09-07
#575: [Examples restyling] Tooltips (closes #575)

### LINKED ISSUE #575: [Examples restyling] Tooltips
← #566

## requirements
We want at least 4 example: 
- default: single text with tooltip
- sizes: two text with tooltip (one 'small', the other 'big')

image

## specs

## misc
{optional: other useful info}

veej:
Issue #575


src/Tooltip/examples/positions.js
```diff
@@ -0,0 +1,44 @@
+class Example extends React.Component {
```
veej: @FrancescoCioria this example has a lot of layout-related noise... Not sure if we want to keep it. What do you think?

FrancescoCioria: why not making this example more interactive by adding a dropdown to let the user change the current position without touching the code?

src/Tooltip/examples/events.js
```diff
@@ -0,0 +1,30 @@
...
+
+    return (
+      <FlexView vAlignContent='center'>
+        <Tooltip popover={{ ...popover, event: 'hover' }}>
+          <span>Hover me!</span>
+        </Tooltip>
+        <Divider size='large' orientation='horizontal' />
+        <Tooltip popover={{ ...popover, isOpen }}>
+          <Button flat size='tiny' onClick={onClick} label='Click me!' />
+        </Tooltip>
```
FrancescoCioria: what about adding also an example of an always open `Tooltip`? Do you think it would be confusing?

veej: IMO the example with `Button` is enough to show the use of `isOpen` prop.
Moreover we could show the "always open" feature in other examples, like `positions` and `sizes`, in order to allow a better comparison of different options.

src/Tooltip/examples/positions.js
veej: evaluating with @alessandravilla... consider that we will have noise due to the dropdowns logic.

PR #580, opened 2016-09-08
#525: LoadingSpinner style should be called loadingSpinner.scss (closes #525)

### LINKED ISSUE #525: LoadingSpinner style should be called loadingSpinner.scss
## description
`LoadingSpinner` style is called `style.scss` instead of `loadingSpinner.scss`

## how to reproduce

## specs
rename it to `loadingSpinner.scss`

## misc
{optional: other useful info}

FrancescoCioria:
Issue #525

### tests performed
`npm run build` --> ✅ 
`npm start` --> Loading spinner example worked as usual


veej:
good 👍 
Needs backporting?

FrancescoCioria:
yes, once the split is over. But fortunately we have a very nice CHANGELOG.md to help us remind of every breaking change :)
(the mid-term goal is to move the responsibility to adapt projects to brc to their own maintainers)
(this change is very simple so no need to worry either)

PR #581, opened 2016-09-08
#549: [KitchenSink] loading time is high (closes #549)

### LINKED ISSUE #549: [KitchenSink] loading time is high
## description
When loading a component page, it may take up to a few seconds to render the page.
After speaking to @veej, this is because we load README files for each component directly from github.

## specs
Since there's no real reason for loading content over the internet, we could just include the READMEs in the build.

/cc @FrancescoCioria @ascariandrea

FrancescoCioria:
Issue #549

### tests performed
- `rm -rf node_modules && npm i && npm start`
  - app builds and starts correctly
  - internal components examples are fine (also example that use `FlexView@1.0.1`)
  - external components example are fine
    - using examples and readmes from github (I know because I updated FlexView examples this morning)
    - using components from github (I know because I updated FlexView examples this morning)


gabro:
@FrancescoCioria check the CI error

gabro:
by the way, I was thinking: can't the showroom just be a separate project within the same repo? So that it can depend on `..` for `react-components` but have its own dependencies

PR #588, opened 2016-09-09
React 15

gabro:
This PR

- updates `devDependencies` to react 15 (showroom builds find, tests pass)
- includes react 0.14 and react 15 in `peerDependencies`

Only doubt I have: we have a dependency on `react-addons-transition-group`, which has a peerDependency on react.

This is problematic: if we update it to version 15, it brings along a react 15 dependency. If we leave it to 0.14, same thing for react 0.14

Long story short: we can't have a `peerDependency` on both react 0.14 and 15 because we're using a dependency that forces us to either one or the other.

Thoughts? @FrancescoCioria @giogonzo


gabro:
maybe something like 

```json
{
  "dependencies": {
    "react-addons-transition-group": "^0.14 || ^15"
  }
}
```

can work?

gabro:
Ok, I took a stab at this (see latest commits).

We know allow both 0.14 and 15 versions of react-addons-transition-group. I'm not sure npm will resolve the version correctly, but at least the current set of ranges shouldn't prevent external projects to install mutually compatible versions (manually if npm fails in doing so)

FrancescoCioria:
👍 👍

PR #601, opened 2016-09-09
#419: Menu: Menu is dismissed when user clicks on disabled menu item (closes #419)

### LINKED ISSUE #419: Menu: Menu is dismissed when user clicks on disabled menu item
## description
- Menu component is dismissed when user clicks on disabled menu item: this is a wrong behavior.

## how to reproduce
- Create a `Menu` with disabled actions

## specs
- Don't call `onClick` callback when action is disabled

## misc
c/c @buildo/frontend

gabro:
Issue #419

### tests performed
unit tests :)

### DIFF
--- a/test/components/Menu.test.js
+++ b/test/components/Menu.test.js
@@ -1,0 +1,41 @@
+import React from 'react';
+// import renderer from 'react-test-renderer';
+import { mount } from 'enzyme';
+import { Menu } from '../../src/DropdownMenu';
+
+const exampleProps = {
+  options: [
+    { title: 'Preferences', type: 'item' },
+    { title: 'Preferences', type: 'item' }
+  ]
+};
+
+describe('Menu', () => {
+
+  // ignored because react-test-renderer cannot co-exist with enzyme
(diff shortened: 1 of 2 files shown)


test/components/Menu.test.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+// import renderer from 'react-test-renderer';
```
gabro: if I don't comment this out, the tests fail. Importing `enzyme` and `react-test-renderer` together doesn't work because of a bug in React that's already fixed in master and will land in react 15.4

test/components/Menu.test.js
```diff
@@ -0,0 +1,42 @@
...
+import { mount } from 'enzyme';
+
```
ascariandrea: Extra line :)

ascariandrea:
Excluded the minor comment code lgtm! 🌸

test/components/Menu.test.js
gabro: thanks, rebased!

PR #606, opened 2016-09-12
#604: Dropdown `onChange` prop should have `value` as argument (closes #604)

### LINKED ISSUE #604: Dropdown `onChange` prop should have `value` as argument
## description
After https://github.com/buildo/react-components/pull/537, `onChange` gets `option` object (= `{ value, ...others }`) instead than just `value`.
See https://github.omnilab.our.buildo.io/buildo/aliniq/issues/4302#issue-5407

## how to reproduce
- http://react-components.buildo.io/#/sections/components/component/time-picker?openSections=%3Bcomponents example is currently broken (try to change time)

## specs
- pass `simpleValue: true` to `Select` (see https://github.com/JedWatson/react-select/blob/master/src/Select.js#L103)

## misc
{optional: other useful info}

francescogior:
Issue #604

### tests performed
- `npm start`
- `#/sections/components/component/time-picker?openSections=%3Bcomponents`
- try to change time
- it shouldn' break

### DIFF
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -123,6 +123,7 @@ export default class Dropdown extends React.Component {
       clearable,
       backspaceRemoves: t.Nil.is(backspaceRemoves) ? clearable : backspaceRemoves,
       resetValue: null,
+      simpleValue: true,
       className: cx('dropdown', className, this.getCustomClassNames()),
       value: this.valueToOption(this.getValue(), options),
       onInputKeyDown,


FrancescoCioria:
@francescogior are Dropdown's examples still working?

francescogior:
Yes, because `Dropdown` supports `value` both as primitive `String|Number` and as option `object`, because of https://github.com/buildo/react-components/blob/master/src/dropdown/Dropdown.js#L75-L85

FrancescoCioria:
👍

PR #621, opened 2016-09-21
#618: [Style Linter] Fix all style files (closes #618)

### LINKED ISSUE #618: [Style Linter] Fix all style files
← #614

## requirements
- after running the npm task to lint style files resolve them
- may we need to tune the `.stylelintrc`to satisfy fully our needs

## specs

## misc
{optional: other useful info}

ascariandrea:
Issue #618

### tests performed
I couldn't run the showroom cause I got an error :/

@FrancescoCioria Just give me feedback on this, we're going to test it before merge.

### DIFF
--- a/src/DropdownMenu/dropdownMenu.scss
+++ b/src/DropdownMenu/dropdownMenu.scss
@@ -1,23 +1,19 @@
 @import '../utils/theme/variables.scss';
-
 $text-color: $brc-coolGrey !default;
 $text-color-hover: $brc-battleshipGrey !default;
-
 $icon-color: $text-color !default;
 $icon-color-hover: $text-color-hover !default;
 
 .dropdown-menu {
   position: relative;
   cursor: pointer;
-
   color: $text-color;
 
   &:hover {
     color: $text-color-hover;
   }
 
-  .dropdown-menu-icon-container { 
-
+  .dropdown-menu-icon-container {
     color: $icon-color;
 
     &:hover {
@@ -25,7 +21,7 @@ $icon-color-hover: $text-color-hover !default;
     }
 
     &.isOpen {
-     z-index: 101;
+      z-index: 101;
     }
   }
 
@@ -35,12 +31,11 @@ $icon-color-hover: $text-color-hover !default;
 
   .dropdown-menu-overlay {
     position: fixed;
-    width: 100%;
-    height: 100%;
     top: 0;
     left: 0;
     z-index: 100;
+    height: 100%;
+    width: 100%;
     background-color: transparent;
   }
-
 }
--- a/src/DropdownMenu/examples/userMenu.scss
+++ b/src/DropdownMenu/examples/userMenu.scss
@@ -1,23 +1,23 @@
 @import '../../utils/theme/variables.scss';
 
 .user-menu {
+  padding: 6px 10px;
   height: 50px;
   min-width: 210px;
-  padding: 6px 10px;
 
   .icon {
     font-size: 7px;
   }
 
   .user-name {
+    color: $brc-darkGrey;
     font-size: 16px;
     font-weight: $brc-semibold;
-    color: $brc-darkGrey;
   }
 
   .user-role {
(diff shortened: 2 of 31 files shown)


FrancescoCioria:
@ascariandrea the order of the properties is very opinionated and IMO needs a full discussion in separate thread. I would remove it from this PR

ascariandrea:
@FrancescoCioria yep, I'm agree but it helps to keep code ordered while we decide a shared order. And if we don't like it we could then change it after easily cause it's fixable.
Anyway I just formatted using the only configuration I found so far and then removed the ordering rules.

We can consider the [zen](https://github.com/csscomb/csscomb.js/blob/dev/config/zen.json) order used by `csscomb` as a starting point.

src/DropdownMenu/dropdownMenu.scss
```diff
@@ -1,31 +1,27 @@
 @import '../utils/theme/variables.scss';
-
 $text-color: $brc-coolGrey !default;
 $text-color-hover: $brc-battleshipGrey !default;
-
```
FrancescoCioria: I possible it would be nice keep this white lines between block of variables

src/DropdownMenu/examples/userMenu.scss
```diff
@@ -1,36 +1,36 @@
...
+    font-size: 14px;
     font-weight: $brc-semibold;
   }
 
   .user-avatar {
     margin-right: 10px;
   }
 
-  & ~ div .menu {
+  ~ div .menu {
```
FrancescoCioria: was the `&` causing this css block to be skipped? we should test the final result of this CSS and see if it changes or not

src/Tooltip/tooltip.scss
```diff
@@ -47,91 +42,109 @@ $arrow-color: $brc-darkGrey !default;
     font-weight: 400;
   }
 
-  &:after, &:before {
-    border: solid transparent;
+  &::after,
+  &::before {
```
FrancescoCioria: is the `::` a rule? why is it useful? if it's only to support very old browser I vote to change this rule to accept a single `:`

src/dropdown/dropdown.scss
```diff
@@ -67,29 +55,33 @@ $height-small: 30px !default;
 }
 
 .Select.dropdown {
-  @include normalize;
-  @include height($height);
   box-sizing: border-box;
+
+  @include normalize();
```
FrancescoCioria: this is the same as writing `@include normalize` right? anyway, we should test this too to see if final result is the same as it was before

src/DropdownMenu/dropdownMenu.scss
ascariandrea: I think the only way is to put a comment between blocks, that makes sense too

src/Tooltip/tooltip.scss
ascariandrea: It's not to support old browsers, is to distinguish the [`pseudo-elements`](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) from the [`pseudo-classes`](https://developer.mozilla.org/en-US/docs/Web/CSS/pseudo-classes).

src/dropdown/dropdown.scss
ascariandrea: Yep, it's the same!

src/DropdownMenu/examples/userMenu.scss
FrancescoCioria: @ascariandrea did you test this?

ascariandrea: Yep, I tested but it was pending.. (the new github workflow confuses me)

src/Tooltip/tooltip.scss
FrancescoCioria: 👍  (I didn't know that)

ascariandrea:
@FrancescoCioria rebased, can we merge this?

PR #626, opened 2016-09-22
#625: Dropdown: $text-color-* variables have no effect (closes #625)

### LINKED ISSUE #625: Dropdown: $text-color-* variables have no effect
## description
overriding the `dropdown` style setting `$text-color-*` variables to a value different from default has no effect

## how to reproduce
override `$text-color-*` variables in dropdown style. Changes have no effect.

## specs
new version of `react-select` is setting the color of `.Select-value-label`, contained in `.Select-control`. Since we're using `$text-color-*` variables to set the color of `.Select-control`, the new `react-select` rules are overriding the desired behavior. 

## misc
{optional: other useful info}

veej:
Issue #625

### tests performed
- change `$text-color-*` values
- every new value should work as expected

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -203,7 +203,10 @@ $height-small: 30px !default;
 
   &:not(.is-disabled):hover {
     background: $background-hover;
-    color: $text-color-hover;
+
+    &.has-value .Select-value .Select-value-label {
+      color: $text-color-hover;
+    }
 
     .Select-placeholder {
       color: $placeholder-color-hover;
@@ -217,7 +220,10 @@ $height-small: 30px !default;
   &.is-open {
     background: $background-open;
     border-color: $border-color-open;
-    color: $text-color-open;
+
+    &:hover > .Select-control .Select-value .Select-value-label {
+      color: $text-color-open;
+    }
 
     &:hover .Select-placeholder, .Select-placeholder {
       color: $placeholder-color-open;
@@ -231,7 +237,10 @@ $height-small: 30px !default;
 
   &.has-value .Select-control {
     padding-right: 45px;
-    color: $text-color-has-value;
+
+    .Select-value .Select-value-label {
+      color: $text-color-has-value;
+    }
 
     .Select-placeholder {
       color: $text-color-has-value;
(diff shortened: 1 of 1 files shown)


src/dropdown/dropdown.scss
```diff
@@ -203,7 +203,10 @@ $height-small: 30px !default;
 
   &:not(.is-disabled):hover {
     background: $background-hover;
-    color: $text-color-hover;
+
+    &.has-value .Select-value .Select-value-label {
+      color: $text-color-hover;
```
FrancescoCioria: wait, this is really different from before: you're using `$text-color-hover` only when it has a value? why?

src/dropdown/dropdown.scss
```diff
@@ -217,7 +220,10 @@ $height-small: 30px !default;
   &.is-open {
     background: $background-open;
     border-color: $border-color-open;
-    color: $text-color-open;
+
+    &:hover > .Select-control .Select-value .Select-value-label {
+      color: $text-color-open;
```
FrancescoCioria: again, this is called `$text-color-open` but you're using only when hovering (as if it was `text-color-open-hover`)

src/dropdown/dropdown.scss
```diff
@@ -231,7 +237,10 @@ $height-small: 30px !default;
 
   &.has-value .Select-control {
     padding-right: 45px;
-    color: $text-color-has-value;
+
+    .Select-value .Select-value-label {
+      color: $text-color-has-value;
```
FrancescoCioria: here you're finally using `text-color-has-value` but it's working only because you added also `.Select-control` otherwise it would be overwritten by the first one

src/dropdown/dropdown.scss
veej: whoops... actually this IS wrong, and that's why I looked for another solution better than using `&.has-value`

src/dropdown/dropdown.scss
veej: new version should be more reliable :P

veej:
@FrancescoCioria this version should be better than the previous one. What do you think?

PR #633, opened 2016-09-27
#612: TimePicker selection, keyboard selection and focus (closes #612)

### LINKED ISSUE #612: TimePicker selection, keyboard selection and focus
NOTE: opening this here because I feel it's related to the recent dropdown upgrade. Some or all of the issues below might be due to wrong customizations on aliniq

## description
dropdown before (hovering first line and then moving with down key):
[https://gyazo.com/230abb2117ffa0b9f92781d843124cda](https://gyazo.com/230abb2117ffa0b9f92781d843124cda)

dropdown after (hovering first row and then moving with down key):
[https://gyazo.com/c1210bf8b8d221a3825533c9c34d1f5d](https://gyazo.com/c1210bf8b8d221a3825533c9c34d1f5d)

It seems we just have an "improvement" here (the currently selected item, if any, is highlighted. The focused item is highlighted in a darker color)

timepicker before (moving with down/up keys):
[https://gyazo.com/03645cea1695b74629c7a52c9081f62e](https://gyazo.com/03645cea1695b74629c7a52c9081f62e)

timepicker after (moving with up/down keys):
[https://gyazo.com/9315e9eaee9487db4c7f36fa4399ff1d](https://gyazo.com/9315e9eaee9487db4c7f36fa4399ff1d)

we have a regression here instead:
- the first line is always highlighted even if another item is selected.
- you can't select items moving with up/down keys

Also, there's no distinction between focused and selected items

## how to reproduce
before: http://alinity-pro-sales.omnilab-development.com
after: http://qia.omnilab-development.com:3000

## specs
{optional: describe a possible fix for this defect, if not obvious}

## misc
{optional: other useful info}

FrancescoCioria:
Issue #612

### tests performed
- Time picker options are computed correctly and refreshed whenever input changes
- Can navigate through options with arrow keys ✅ 

[https://gyazo.com/7250d349efab16bc86a488990a4500dc](https://gyazo.com/7250d349efab16bc86a488990a4500dc)

### DIFF
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -5,8 +5,6 @@ import Dropdown from '../dropdown/Dropdown';
 import range from 'lodash/range';
 import flatten from 'lodash/flatten';
 
-const options = [];
-
 export const H24 = '24h';
 export const H12 = '12h';
 export const inputError = {};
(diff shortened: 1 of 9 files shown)


FrancescoCioria:
for future reviewer: `options` was always empty before (`[]`) --> [this logic](https://github.com/buildo/react-components/blob/master/src/dropdown/Dropdown.js#L105) stopped any input key event.

PR #636, opened 2016-09-28
#635: Dropdown: add sass variable $option-background-selected (closes #635)

### LINKED ISSUE #635: Dropdown: add sass variable $option-background-selected
## description
the new default style from `react-style@1.0.0-rc1` sets a different background color for the selected option.
This is not visible in the showroom because we're accidentally overwriting it with stronger css selectors but it's clearly visible in other projects (ex: QIA) where it's not easily customizable through variables.

The concept of having a different background color for the selected option is legitimate so we should add a new variable to let final users decide which color they want.

## how to reproduce

## specs
add `$option-background-selected` and use it when `is-selected`

## misc
{optional: other useful info}

FrancescoCioria:
Issue #635

### tests performed
- set `option-background-selected` to "green"
- set `option-color-selected`  "red"

-->
[https://gyazo.com/b61cf1b470971dfa28bef31b3bc56420](https://gyazo.com/b61cf1b470971dfa28bef31b3bc56420)

PS: I also tested that option-background-hover still works:
- set `$option-background-hover` to "black"
  - option turns black when hovered

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -24,7 +24,9 @@ $menu-background: $brc-white !default;
 $menu-border-color: $brc-silver !default;
 $option-color: rgba($brc-darkGrey, .5) !default;
 $option-color-hover: $brc-darkGrey !default;
+$option-color-selected: $option-color !default;
 $option-background-hover: rgba($brc-azure, .15) !default;
+$option-background-selected: $menu-background !default;
 $multiple-value-color: #007eff !default;
 $multiple-value-border-color: $multiple-value-color !default;
 $multiple-value-background-color: #daecf6 !default;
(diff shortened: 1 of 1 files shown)


src/dropdown/dropdown.scss
```diff
@@ -24,7 +24,9 @@ $menu-background: $brc-white !default;
 $menu-border-color: $brc-silver !default;
 $option-color: rgba($brc-darkGrey, .5) !default;
 $option-color-hover: $brc-darkGrey !default;
+$option-color-selected: $option-color !default;
 $option-background-hover: rgba($brc-azure, .15) !default;
+$option-background-selected: $menu-background !default;
```
FrancescoCioria: to avoid making this issue "breaking" I set `option-color-selected` and `option-background-selected` to the defaults (this is what we were getting before v1.0.0-rc1 and is what we still expect in QIA)

PR #645, opened 2016-10-04
Automatically detect if breaking or patch release

FrancescoCioria:
## TestPlan
- comment every command after `npm version ...` (I don't want to release a version on `npm` :) )
- run on branch
  - throws error because not on "master"
- run
  - throws error because there's nothing to release ✅ 
- run and take second tag
  - releases PATCH version ✅ 
- run and take fifth tag
  - releases BREAKING version ✅


releaseVersion.js
```diff
@@ -0,0 +1,80 @@
...
+    throw new Error('Your local and remote branches have diverged. Please put them in sync. Aborting.');
+  }
+
+  const BASE_URL = 'https://api.github.com/repos/buildo/react-components';
+
+  const isVersionTag = tag => {
+    return _.startsWith(tag.name, 'v') && _.every(tag.name.slice(1).split('.'), s => typeof parseInt(s) === 'number');
+  };
+
+  axios.get(`${BASE_URL}/tags`).then(res => {
```
gabro: any particular reason for using `axios` instead of `octokat`?

FrancescoCioria: Only that `axios` is already a dev dependency of `b-r-c`

gabro: I thought so. Then 👍

PR #646, opened 2016-10-10
use config from stylelint-config-buildo

FrancescoCioria:


.stylelintrc
```diff
@@ -1,23 +1,3 @@
...
-    "scss/dollar-variable-colon-space-after": "always",
-    "scss/dollar-variable-colon-space-before": "never",
-    "scss/dollar-variable-no-missing-interpolation": true,
-    "scss/double-slash-comment-empty-line-before": "always",
-    "scss/media-feature-value-dollar-variable": "always",
-    "scss/operator-no-unspaced": true,
-    "scss/selector-no-redundant-nesting-selector": true
-
-  }
+  "extends": "stylelint-config-buildo"
```
gabro: `extends: buildo`?

FrancescoCioria: not supported :(

gabro: ¯\_(ツ)_/¯

PR #648, opened 2016-10-12
Add yarn compatibility

gabro:
## test plan

``` sh
rm -rf node_modules
yarn
```

``` sh
rm -rf node_modules
npm i
```

👍


ghDepsInstall.sh
```diff
@@ -1,5 +1,6 @@
 #! /bin/bash
 
 if [[ $(git config --get remote.origin.url) == *buildo/react-components.git* ]]; then
-  npm i ./ghDeps;
+  # use yarn if installed, or use npm instead
+  command -v yarn > /dev/null 2>&1 && yarn add ./ghDeps || npm i ./ghDeps
```
giogonzo: as I understand it, `yarn add` is the equivalent of `npm i --save` (not just `npm i`). Aren't we adding more deps to the top-level package here (eg: `"ghDeps": "file://./ghDeps"`)? Or does it work differently in this case?

gabro: right, I thought about it but after `yarn` I couldn't see any diff...

gabro: shit, ok, yarn doesn't run `postinstall` after `yarn`. I'll investigate

gabro: https://github.com/yarnpkg/yarn/pull/800

FrancescoCioria:
@gabro is this PR just a test?
If not, (when you have time) can you also open an issue about `yarn` where you briefly describe why you'd like to support it? (it looks cool -> If we're serious about it I'd like to give it more visibility across people)

gabro:
Closing for now

PR #654, opened 2016-10-31
#651: New tablo component (closes #651)

### LINKED ISSUE #651: New tablo component
## requirements

The new `Tablo` component will replace the `Table` wrapper in the next future.

`Tablo` is essentially a cleaned and improved version of `Table`, based on [`fixed-data-table-2`](https://github.com/schrodinger/fixed-data-table-2), with new, modular, features (i.e. shortcuts, autosize, columns reordering, sorting, ecc.) added around `Table` during the development of our projects.
## specs

## misc

{optional: other useful info}

veej:
Issue #651

### tests performed
- look at the shiny examples 🌟


src/tablo/examples/scrollable.js
```diff
@@ -0,0 +1,94 @@
...
...ncidunt.com",phone:"(867) 528-8896",company:"Interdum PC"},{name:"Zenaida I. Booth",email:"Fusce.aliquam.enim@amet.org",phone:"(737) 462-1009",company:"Adipiscing Inc."},{name:"Vance E. Berg",email:"at@sociisnatoquepenatibus.co.uk",phone:"(630) 858-4328",company:"Rutrum Eu Ultrices LLP"},{name:"Norman Shaw",email:"Nunc.laoreet.lectus@nislQuisque.co.uk",phone:"(990) 752-4673",company:"Suspendisse Ac Industries"},{name:"Thaddeus Anthony",email:"semper.et@magnaa.ca",phone:"(106) 259-3496",company:"Accumsan LLP"},{name:"Eagan Y. Hanson",email:"mi@necligulaconsectetuer.com",phone:"(917) 135-9190",company:"Ultrices Foundation"},{name:"Hayes M. Holloway",email:"mi.Aliquam.gravida@odio.ca",phone:"(358) 977-1127",company:"Mauris Eu Elit LLC"},{name:"Isabelle V. Hammond",email:"vel.quam@inconsectetueripsum.co.uk",phone:"(495) 896-0515",company:"Donec LLP"},{name:"Nicholas N. Simon",email:"non@hendreritaarcu.edu",phone:"(890) 239-3187",company:"Ut Pellentesque Eget Corporation"},{name:"Rae Q. Alvarado",email:"dapibus@acturpis.ca",phone:"(277) 280-5712",company:"Eget Industries"},{name:"Ebony Farmer",email:"sem.mollis.dui@Nuncpulvinar.org",phone:"(951) 583-6526",company:"Nec Mollis Company"}];
```
veej: Is it worth to replace this with a random data generator?

showroom/components/App.js
```diff
@@ -17,7 +18,7 @@ const scope = {
   React, ReactDOM,
   t,
   log: x => console.log(x), // eslint-disable-line no-console
-  find, partial, reject,
+  find, partial, reject, sortBy,
```
veej: Used to sort data in the first example

src/index.js
```diff
@@ -28,3 +28,4 @@ export BrowserDetector from './browser-detector';
 export TimePicker from './time-picker';
 export FocusableView from './focusable-view';
 export ResizeSensor from './resize-sensor';
+export Tablo, { Column as TabloColumn, Cell, Header, Footer } from './tablo';
```
veej: I had to rename `Column` to `TabloColumn` to avoid name collision with `Table` component.
Not so nice, as example could be confusing (`import { TabloColumn } from '...'` won't work)

src/tablo/Tablo.js
```diff
@@ -0,0 +1,117 @@
...
+  footerHeight: t.Number,
+  onRowMouseEnter: maybe(t.Function),
+  onRowMouseLeave: maybe(t.Function),
+  scrollLeft: maybe(t.Integer),
+  scrollTop: maybe(t.Integer),
+  onScrollStart: maybe(t.Function),
+  onScrollEnd: maybe(t.Function),
+  children: t.ReactChildren,
+
+  // private
```
veej: These props are used internally for plugin functionalities. Should not be overwritten by devs, or plugin functionalities will break (maybe we should filter them in plugins)

src/tablo/index.js
```diff
@@ -0,0 +1,10 @@
+import Tablo from './Tablo';
+import { autosize, columnsResize, columnsReorder, scrollable, selectable, sortable } from './plugins';
+
+export ColumnGroup from './ColumnGroup';
+export Column from './Column';
+export Cell from './Cell';
+export Header from './Header';
+export Footer from './Footer';
+
+export default autosize(columnsResize(columnsReorder(scrollable(selectable(sortable(Tablo))))));
```
veej: I'm using decorators here to avoid issues with `react-readme-generator`.

src/tablo/plugins/autosize/autosizeGrid.js
```diff
@@ -0,0 +1,65 @@
...
+import cx from 'classnames';
+import FlexView from 'react-flexview';
+import ResizeSensor from '../../../resize-sensor/ResizeSensor';
+
+export default (Grid) =>
+
+  @props({
+    className: maybe(t.String),
+    autosize: maybe(t.Boolean)
+  }, { strict: false })
```
veej: we can safely use `{ strict: false }` as additional props will be checked by `Tablo` component

package.json
```diff
@@ -38,11 +38,15 @@
     "classnames": "^2.2.3",
     "css-element-queries": "^0.3.2",
     "debug": "^2.2.0",
+    "faker": "^3.1.0",
```
veej: Used to generate fake rows in examples

src/tablo/Tablo.js
```diff
@@ -0,0 +1,117 @@
...
+ * @param selectedRows - the list of selected row ids
+ * @param onRowsSelect - callback to be called when the currently selected rows change
+ * @param selectionType - single = only one selected row at a time, multi = multiple selection, none = selection disabled
+ * @param hoveredRowIndex - the id of the hovered row
+ * @param onHoverRowChange - callback to be called when the hovered row changes
+ * @param sortBy - id of the column according which the data should be ordered
+ * @param sortDir - sorting direction
+ * @param onSortChange - callback to be called when sorting change
+ *
+ * @param scrollToRow - Private
```
veej: We should filter 'Private' props in doc generator.

src/tablo/plugins/columnsReorder/columnsReorderGrid.js
```diff
@@ -0,0 +1,119 @@
...
+    columnsOrder: maybe(list(t.String)),
+    onColumnsReorder: maybe(t.Function)
+  }, { strict: false })
+  @pure
+  @skinnable(contains(Grid))
+  class ColumnsReorderGrid extends React.Component {
+
+    constructor() {
+      super();
+      this.uniqueId = uniqueId('tablo_');
```
veej: This uniqueId is used to avoid d&d between different tables

src/tablo/Tablo.js
```diff
@@ -38,10 +39,15 @@ export const defaultWidth = 500;
   groupHeaderHeight: t.Number,
   footerHeight: t.Number,
   children: t.ReactChildren,
+  onRowMouseEnter: maybe(t.Function),
```
FrancescoCioria: as you already know this is super-heavy from a performance point-of-view

src/tablo/Column/Column.js
```diff
@@ -16,6 +16,7 @@ const argsTypes = struct({
   data: maybe(t.Array),
   name: t.String,
   fixed: maybe(t.Boolean),
+  flexGrow: maybe(t.Number),
```
FrancescoCioria: we should consider making this a `union([t.Boolean, t.Number])` to make it work like `FlexView` (`flexGrow={true} ---> flexGrow: 1`)

src/tablo/plugins/columnsReorder/columnsReorder.js
```diff
@@ -28,6 +31,11 @@ export default (Grid) =>
   @skinnable(contains(Grid))
   class ColumnsReorderGrid extends React.Component {
 
+    constructor() {
+      super();
+      this.uniqueId = uniqueId('tablo_');
```
FrancescoCioria: you should be able to define this outside the `constructor`:
```js
uniqueId = uniqueId('tablo_')
```

src/tablo/examples/scrollable.js
FrancescoCioria: yessss :)

src/tablo/plugins/scrollable/scrollable.js
```diff
@@ -0,0 +1,43 @@
...
+  @pure
+  @skinnable(contains(Grid))
+  class ScrollableGrid extends React.Component {
+
+    state = {
+      scrollTop: this.props.scrollTop
+    }
+
+    componentWillReceiveProps(nextProps) {
+      const scrollTopDidntChanged = nextProps.scrollTop === this.props.scrollTop;
```
FrancescoCioria: to make functions more readable we usually pick every needed variable from `props`/`state` in the first line of the function:

```js
const { scrollTop, selectedRows } = this.props
```

(is it defined somewhere in our guidelines? not sure...)

package.json
FrancescoCioria: yaaay!

src/tablo/Footer/footer.scss
```diff
@@ -3,7 +3,7 @@
 
 // no fixed-data-table css classes in subcomponents
 
-.grid {
+.tablo {
   .grid-footer {
```
FrancescoCioria: why didn't you change also this and the other `grid-header`, `grid-x`...?

src/tablo/Cell/Cell.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import t, { maybe, enums, union } from 'tcomb';
+import { props  } from 'tcomb-react';
```
francescogior: I'm fixing it now, this was *quite hard* in a way.

Why don't we have roots modules in brc?

src/tablo/Footer/footer.scss
francescogior: you're right, lex do it!

src/tablo/Tablo.js
```diff
@@ -0,0 +1,81 @@
...
+  rowHeight: t.Number,
+  headerHeight: t.Number,
+  groupHeaderHeight: t.Number,
+  footerHeight: t.Number,
+  children: t.ReactChildren
+})
+export default class Tablo extends React.Component {
+
+  static defaultProps = {
+    width: defaultWidth,
```
francescogior: I partially agree but I'd like to discuss it more in depth

francescogior: While I agree to remove `width` and `height` as default props (they are actually totally random) I don't think the default props are to be avoided as they are evil.
For instance, for me a default row height makes sense in a table component and I'd blame the dev if he forgets to pass the prop explicitly when required by specs.

Anyway what happens here is that actually `width` and `height` are ignored unless one doesn't pass `autosize={false}` (autosizing future turned out to be extremely helpful in all the usages of the tables in our projects)

PR #664, opened 2016-11-15
#663: TimePicker missing "is-selected" class (closes #663)

### LINKED ISSUE #663: TimePicker missing "is-selected" class
## description
the selected option in the TimePicker component doesn't have the "is-selected" class

## how to reproduce
inspect the selected option using chrome devTools 
Note: to inspect an option while the drop-menu is opened use the "cmd+F8+\" shortcut 

## specs

## misc
{optional: other useful info}

sawsanias:
Issue #663

### tests performed
Start the react-component application
go to [http://localhost:8080/#/sections/components/component/time-picker](url)
select one of the options from the drop-down menu of the time-picker (either 12H or 24H format)
inspect the option of the selected time (to keep the menu opened while the devtool is opened click `cmd+F8+\`) 
The option should have the "is-selected" class

image

### DIFF
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -189,9 +189,7 @@ export default class TimePicker extends React.Component {
       ...props
     } = this.props;
 
-    const value = userValue ? toOption({
-      ...userValue, timeFormat
-    }) : undefined;
+    const value = userValue ? formatTime24(userValue) : undefined;
 
     return {
       ...props,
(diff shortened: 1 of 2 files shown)


francescogior:
@sawsanias please fix tests!

FrancescoCioria:
@sawsanias poor nemobot has reminded you 4 times to add a test plan and it's still missing... Please add one **before** assigning for a review :)

PR #666, opened 2016-11-16
#665: Update eslint-config-buildo@^0.2.0 (closes #665)

### LINKED ISSUE #665: Update eslint-config-buildo@^0.2.0
## requirements
- replaced `eslint-config-buildo` git master dependency with semver `^0.2.0`

## specs

## misc
{optional: other useful info}

ascariandrea:
Issue #665

### tests performed
- `npm run lint` ✔️ 


gabro:
did we discuss this? Why are we depending on a version and not on master?

FrancescoCioria:
> did we discuss this? Why are we depending on a version and not on master?

nope. Personally, for a project like `brc` I'd like to keep the version on `master` as, if a deploy fails (still very very rare) no one is gonna die and I can simply fix it in the following hours or even days :)

ascariandrea:
What's the pro of depending on `master`?
We have *greenkeeper* on this repo and the `eslint-config-buildo` should be versioned for other projects.

PR #670, opened 2016-11-17
Re-enable previously broken tests with react 15.4

gabro:
A few tests were disabled because of an issue now fixed in React 15.4.

This PR updates React and React-related dependencies to 15.4 and takes advantage of the fixes to re-enabled and/or fix various tests.

### test performed
unit tests (duh!)


src/background-dimmer/BackgroundDimmer.js
```diff
@@ -77,8 +78,7 @@ export default class BackgroundDimmer extends React.Component {
 
     const fixedStyle = { position: 'fixed', top: 0, left: 0, right: 0, bottom: 0 };
     return {
-      ...props,
-      stopPropagation,
+      ...omit(props, ['onClickOutside', 'stopPropagation']),
```
FrancescoCioria: `stopPropagation` is not a prop so it shouldn't be necessary to omit it

gabro: oh, it was more broken than I thought. See the deleted line! I'll fix it

PR #687, opened 2016-11-29
#686: Add tests to Icon (closes #686)

### LINKED ISSUE #686: Add tests to Icon
## requirements
Icon is testless

## specs
Add unit tests

gabro:
Issue #686

### tests performed
unit tests

### DIFF
--- a/src/Icon/Icon.js
+++ b/src/Icon/Icon.js
@@ -30,10 +30,10 @@ export default class Icon extends React.Component {
     onClick: () => {}
   };
 
-  getLocals() {
-    const { className, icon, color, style = {} } = this.props;
+  getLocals(props) {
+    const { className, icon, color, style = {} } = props;
     return {
-      ...this.props,
+      ...props,
       className: cx('icon', `icon-${icon}`, className),
       style: { ...style, color: color || style.color }
     };
--- a/src/Icon/Icon.js
+++ b/src/Icon/Icon.js
@@ -31,14 +31,17 @@ export default class Icon extends React.Component {
   };
 
   getLocals() {
+    const { className, icon, color, style = {} } = this.props;
     return {
-      ...this.props
+      ...this.props,
+      className: cx('icon', `icon-${icon}`, className),
+      style: { ...style, color: color || style.color }
     };
   }
 
-  template({ icon, color, className, style = {}, onClick, paths }) {
+  template({ icon, className, style, onClick, paths }) {
     return icon ? (
-      <i className={cx('icon', `icon-${icon}`, className)} style={{ ...style, color: color || style.color }} onClick={onClick}>
+      <i className={className} style={style} onClick={onClick}>
         {paths > 1 && range(paths).map(k => <span className={`path${k + 1}`} key={k} /> )}
       </i>
     )
--- a/test/components/Icon.test.js
+++ b/test/components/Icon.test.js
@@ -1,0 +1,100 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { shallow } from 'enzyme';
+
+import Icon from '../../src/Icon';
+
+describe('Icon', () => {
+
+  it('renders correctly', () => {
+    const component = renderer.create(
+      <Icon icon='foo' />
+    );
+    expect(component).toMatchSnapshot();
+  });
+
+  it('renders null if no icon is passed', () => {
+    const component = renderer.create(
+      <Icon />
+    );
+    expect(component.toJSON()).toBeNull();
+  });
+
+  it('renders no child when path is 1', () => {
+    const component = shallow(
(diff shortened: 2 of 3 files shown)


test/components/Icon.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+import Icon from '../../src/Icon';
+
+describe('Icon', () => {
+
+  it('renders correctly', () => {
+    const component = renderer.create(
+      <Icon icon='foo' />
+    );
+    expect(component).toMatchSnapshot();
```
dhinus: This seems to work, but all the [official](http://facebook.github.io/jest/blog/2016/07/27/jest-14.html) [examples](https://facebook.github.io/jest/docs/tutorial-react.html) call `toJSON()` before calling `toMatchSnapshot()`.

test/components/Icon.test.js
```diff
@@ -0,0 +1,100 @@
...
+        style: { color: 'blue' },
+        color: 'red'
+      });
+      const { style: { color } = {} } = icon.getLocals();
+      expect(color).toBe('red');
+    });
+
+  });
+
+});
```
dhinus: Don't we usually have a newline at the end of the file? This would be the `eol-last` rule in ESLint, but it's not currently part of our config.

test/components/Icon.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+  it('renders null if no icon is passed', () => {
+    const component = renderer.create(
+      <Icon />
+    );
+    expect(component.toJSON()).toBeNull();
+  });
+
+  it('renders no child when path is 1', () => {
+    const component = shallow(
```
dhinus: With a small change in the expectation, this could be tested using `renderer.create` instead of `shallow`, but I didn't see any performance improvement, and we cannot remove the `enzyme` dependency because we still need it for the `onClick` test below.

test/components/Icon.test.js
```diff
@@ -0,0 +1,100 @@
...
+      <Icon icon='foo' paths={1} />
+    );
+    expect(component.children().length).toBe(0);
+  });
+
+  it('renders a child for each path', () => {
+    const component = shallow(
```
dhinus: Same as above (`renderer.create` vs `shallow`).

src/Icon/Icon.js
```diff
@@ -31,14 +31,17 @@ export default class Icon extends React.Component {
   };
 
   getLocals() {
+    const { className, icon, color, style = {} } = this.props;
```
dhinus: @FrancescoCioria suggested this alternative:
```
getLocals({ className, icon, color, style = {} }) {}
```

OTOH, The Book:tm: [does not agree](https://buildo.gitbooks.io/book/content/frontend/tutorial/step_3.html).

test/components/Icon.test.js
gabro: I plan to ditch react-test-renderer in favor of enzyme in the near future (using https://github.com/adriantoine/enzyme-to-json for seamless serialization), so I think this can stay.

dhinus: 👌

test/components/Icon.test.js
gabro: my authoritative source for this is: https://twitter.com/VilleImmonen/status/772488155584983041

test/components/Icon.test.js
gabro: out of scope?

test/components/Icon.test.js
gabro: same as above :)

src/Icon/Icon.js
gabro: let's stick to the book for now

FrancescoCioria: mmmm ok, but I guarantee you we're doing `getLocals({ ... })` in 90%+ of components in our projects :)

test/components/Icon.test.js
FrancescoCioria: yes, probably out-of-scope (even if `atom` automatically adds a new line which means every file has a new line at the end -> we should probably add this rule)

src/Icon/Icon.js
gabro: are we? Ok, I can change this then and then change the book!

test/components/Icon.test.js
dhinus: Fair enough :grin:

src/Icon/Icon.js
gabro: @dhinus done, pushed and rebased on master.

gabro:
related: https://github.com/buildo/book/pull/6/files

/cc @FrancescoCioria

dhinus:
🎉

PR #693, opened 2016-12-05
#692: use `smooth-release` (closes #692)

### LINKED ISSUE #692: use `smooth-release`
## requirements
replace custom version/publish scripts with `smooth-release`

## specs
install `smooth-release` as a dev dep and use it in release script and changelog step in CI

## misc
{optional: other useful info}

FrancescoCioria:
Issue #692

### tests performed
- npm run release-version -> ti works
- change teomporarily `gh-release` CI step to run in every branch
  - it runs correctly on CI and fails because release already exists on github for current tag
- change teomporarily `changelog` CI step to run in every branch
  - it runs correctly on CI and fails because CHANGELOG.md hasn't changed

didn't test the `push` part of `changelog` step


FrancescoCioria:
> we run CI with --no-validations because we are 100% sure we just fetched the latest from repo?

The reason I turned it off is because on the CI the git work directory is always dirty and validations always fail... anyway, both steps are run only on `master` branch and after a `git pull --ff-only origin master` so they should be pretty safe.

> we are not running publish (npm) on CI right?

no, we're not running it, but just to make it more clear and future-bugs-proof I'll add an explicit `--no-npm-publish` :)

giogonzo:
:+1:

PR #702, opened 2016-12-16
#701: Popover: delay prop validation always fails if you pass an object (closes #701)

### LINKED ISSUE #701: Popover: delay prop validation always fails if you pass an object
## description
The prop `delay: { whenOpen: 100 }` should be valid, but tcomb throws anyway
image

## how to reproduce

## specs
`union` + `struct` = 💣 
replace `struct` with `t.interface`
## misc
{optional: other useful info}

FrancescoCioria:
Issue #701

### tests performed
- warning is no longer there :)
- pass `delay={{ whenOpen: 100, ciao: true }}`
  ~~- tcomb complains (`interface` is indeed `strict`)~~

~~PS: We were using `t.interface` only in `Modal`:~~
image

~~Also in this case it's correct to make it `strict`~~

### DIFF
--- a/src/popover/Popover.js
+++ b/src/popover/Popover.js
@@ -29,7 +29,7 @@ export const Props = {
     isOpen: t.maybe(t.Boolean),
     delay: t.maybe(t.union([
       t.Integer,
-      t.struct({ whenClosed: t.maybe(t.Integer), whenOpen: t.maybe(t.Integer) })
+      t.interface({ whenClosed: t.maybe(t.Integer), whenOpen: t.maybe(t.Integer) })
     ]))
   }),
   id: t.maybe(t.String),


FrancescoCioria:
@veej I think that was the right approach before tomb added t. Interface. Now actually we should always use interface and use struct only for objects out of client control like API results etc. (I didn't change every struct to keep this PR as little as possible). 

@giogonzo do you confirm or should I follow the approach explained in the link?

giogonzo:
yes I think `interface` is easier (we don't loose anything but we spare some boilerplatish code).

`strict` is a separate issue though: `struct`, same as `interface`, is non-strict by default. So probably not in scope for this PR?

FrancescoCioria:
> `strict` is a separate issue though: `struct`, same as `interface`, is non-strict by default. So probably not in scope for this PR?

yeah, it's not in scope you're right. I'll fix it in another issue where I'll switch completely from `struct` to `interface`.

I'll remove the `t.interface.strict = true` code from this PR

PR #717, opened 2016-12-21
#714: [Tablo Perf] handle hover style for rows in CSS (closes #714)

### LINKED ISSUE #714: [Tablo Perf] handle hover style for rows in CSS
← #711

## requirements
`onRowMouseEnter/Leave` callbacks are heavy form a perf point-of-view as they're called twice for each row when users scroll with the mouse pointer above `Tablo`.

As they're used mostly (always so far) to change the background color of a row when hovered, we should handle this simple case in the CSS so to ensure users use them only when really necessary.

## specs
- change the background color of a row when hovered in the CSS
- add a sass variable for it

## misc
{optional: other useful info}

FrancescoCioria:
Issue #714

### tests performed
- hover a row in `default` example
  - row background is colored correctly
- select a row and hover it
  - row background is set to the hover one (hover beats select)
- pass `onRowMouseEnter` and `onRowMouseLeave` to `Tablo`
  - correctly called when entering/leaving a row

### DIFF
--- a/src/tablo/API.md
+++ b/src/tablo/API.md
@@ -13,8 +13,6 @@ A table component based on fixed-data-table-2
 | **selectionType** | <code>multi &#124; single &#124; none</code> |  <code>none</code> | *optional*.
 | **selectedRows** | <code>Array(Integer)</code> | | *optional*. List of selected rows' indexes
 | **onRowsSelect** | <code>Function</code> | | *optional*. Callback to handle <code>selectedRows</code> prop. An array with the indexes of selected rows is passed as parameter
-| **hoveredRowIndex** | <code>Integer</code> | | *optional*.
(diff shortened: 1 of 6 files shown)


FrancescoCioria:
because selected color is applied only if row is not hovered:
```scss
&.selected:not(:hover) .public_fixedDataTableCell_main {
```

so yeah, it doesn't beat it, it's selected which is disabled when row is hovered

francescogior:
Yes sure! But I was discussing
- why one should beat the other one by default (before it was the opposite I guess, and it'd apply as well)
- above all, how to make the user able to customize it?

francescogior:
As discussed offline, using a sass variable `hovered-seelexted-row` (default set to `hovered-row` or `selected-row`)or something may be a good idea

FrancescoCioria:
@francescogior done. TestPlan remained valid after the change

PR #718, opened 2016-12-22
#712: [Tablo Perf] disable prop validation for fixed-data-table-2 also in development (closes #712)

### LINKED ISSUE #712: [Tablo Perf] disable prop validation for fixed-data-table-2 also in development
← #711

## requirements
as done for `fixed-data-table` in `patch-fixed-data-table.js` we should somehow disable prop validations for every `fixed-data-table-2` we're using so to increase considerably `Tablo` speed in development

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Issue #712

### tests performed
mmmm....eeeehm...boooh
I found out the heaviest `propTypes` were actually already disabled in dev by default. I turned off all of them just to be sure.


francescogior:
👍

PR #722, opened 2016-12-30
#721: Panel: header height should be configurable (closes #721)

### LINKED ISSUE #721: Panel: header height should be configurable
## requirements
We would like to have different sizes for panel height. 
The required size should be configurable via props.

## specs
- Add `size` key to the `header` prop. 
  - it can take 3 different values: tiny (40px), small (48px), medium (56px)
  - small should be the default value

## misc
{optional: other useful info}

veej:
Issue #721

### tests performed
- if size is not specified, or `size='small'`, header height is 48px
image

- adding size = `medium`, header height is 56px
image

- adding `size='small'`, header height is 40px
image

### DIFF
--- a/src/Panel/Panel.js
+++ b/src/Panel/Panel.js
@@ -149,7 +149,7 @@ export default class Panel extends React.Component {
       header ?
         <PanelHeader
           title={header.hideTitleWhenExpanded && isExpanded ? undefined : header.title}
-          size={header.size || HeaderSize('small')}
+          size={header.size}
           content={header.content}
           menu={header.menu}
           collapse={header.collapse ? {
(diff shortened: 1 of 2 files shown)


src/Panel/Panel.js
```diff
@@ -148,6 +149,7 @@ export default class Panel extends React.Component {
       header ?
         <PanelHeader
           title={header.hideTitleWhenExpanded && isExpanded ? undefined : header.title}
+          size={header.size || HeaderSize('small')}
```
FrancescoCioria: `small` should be moved to the `defaultProps` (so to have it documented in the README)

veej: Not really sure about this.
It should be moved inside `defaultProps` of `PanelHeader`, and it won't affect the README.

FrancescoCioria: Ok, choose by yourself in which `defaultProps` to put it, but I think this default should be defined in `defaultProps` as we're doing for the other props (I personally think it makes the component more readable as there is a standard place just below `@props` to see every eventual default)

PR #723, opened 2016-12-30
#671: Modal renders header/footer even when not necessary (closes #671)

### LINKED ISSUE #671: Modal renders header/footer even when not necessary
## description
`Modal` always render `header` and `footer` even when content for them is passed.

## how to reproduce
- don't pass `footer` prop
  - footer template is rendered anyway
- don't pass `title` nor `iconClose`
  - header template is rendered anyway

## specs
- render `footer` only if `props.footer` exists
- render `header` only if at least one `props.title` and `props.iconClose` exist

## misc
{optional: other useful info}

ascariandrea:
Issue #671

### tests performed
Added 2 examples for test purpose:
 - modal with no header
 - modal with no footer

Shouldn't this be breaking?

### DIFF
--- a/showroom/components.json
+++ b/showroom/components.json
@@ -205,6 +205,14 @@
         "examples": [
           {
             "url": "modal/examples/default.js"
+          },
+          {
+            "title": "Modal without header",
+            "url": "modal/examples/noHeader.js"
+          },
+          {
+            "title": "Modal without footer",
+            "url": "modal/examples/noFooter.js"
           }
         ]
       },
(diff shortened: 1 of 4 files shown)


FrancescoCioria:
> Shouldn't this be breaking?

In theory no as this the correct behavior described in the mockups: this PR is solving a defect/bug (and solving a bug is always in a way breaking)

Still, as it's been here for months maybe we'd better ensure people acknowledge this issue... I'll add the label `breaking` :)

PR #751, opened 2017-02-07
#749: Tablo: fix drag and drop appearance when using TextOverflow in Header (closes #749)

### LINKED ISSUE #749: Tablo: fix drag and drop appearance when using TextOverflow in Header
## description
- using `TextOverflow` in `Header` causes the `DragPreview` to include all the columns' headers next to the dragged one

## specs
- add `style: transform3d(0,0,0)` to the element wrapping the dragged element (Header is fine)

francescogior:
Issue #749

### DIFF
--- a/src/tablo/Header/Header.js
+++ b/src/tablo/Header/Header.js
@@ -31,6 +31,7 @@ const getLocals = ({
 const template = ({ onClick, className, children }) => {
   return (
     <FlexView
+      style={{ transform: 'translate3d(0, 0, 0)' }}
       className={cx('tablo-header', className)}
       height='100%'
       width='100%'


FrancescoCioria:
@francescogior did you test this in other browsers?
I'm asking cause I think `transform` needs to be prefixed for IE, some versions of firefox and webkit:
https://css-tricks.com/almanac/properties/t/transform/#article-header-id-12

francescogior:
This doesn't actually work, in general:
- it works in our internal project (but it can be fixed there without touching the library)
- it doesn't work in the showroom (adding the `TextOverflow` in the `Header`s)

I'd close this, leaving open the issue for future developments

PR #759, opened 2017-02-14
#758: Tablo: Cell children shoul take the entire row as second argument (closes #758)

### LINKED ISSUE #758: Tablo: Cell children shoul take the entire row as second argument
## description

When data are not normalized to be displayed in a table, a Cell should be able to render its content getting data from different props

## specs
The first argument of Cell children is `data[rowIndex][columnKey]`.
- Pass `data[rowIndex]` as second argument.
- Pass `rowIndex` as third argument

francescogior:
Issue #758

### tests performed
- updated examples as demo

### DIFF
--- a/src/tablo/Cell/Cell.js
+++ b/src/tablo/Cell/Cell.js
@@ -7,6 +7,8 @@ const { maybe, enums, union } = t;
 
 const propsTypes = {
   data: t.Any,
+  rowData: t.Any,
+  rowIndex: maybe(t.Integer),
   children: union([t.Function, t.ReactChildren]),
   backgroundColor: maybe(t.String),
   color: maybe(t.String),
@@ -17,6 +19,8 @@ const propsTypes = {
 
 const template = ({
   data,
+  rowData,
+  rowIndex,
   children,
   backgroundColor,
   color,
@@ -33,7 +37,7 @@ const template = ({
         hAlignContent={hAlignContent}
       >
         <FlexView className='content' vAlignContent='center'>
-          {t.Function.is(children) ? children(data) : children}
+          {t.Function.is(children) ? children(data, rowData, rowIndex) : children}
         </FlexView>
       </FlexView>
     </CellFDT>
--- a/src/tablo/Column/Column.js
+++ b/src/tablo/Column/Column.js
@@ -41,7 +41,9 @@ const Column = (args) => {
 
   const cell = ({ rowIndex, columnKey }) => {
     const elem = find([].concat(children), child => child.type === Cell) || defaultCell;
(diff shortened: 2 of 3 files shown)


src/tablo/Cell/Cell.js
```diff
@@ -7,6 +7,8 @@ const { maybe, enums, union } = t;
 
 const propsTypes = {
   data: t.Any,
+  rowData: t.Any,
```
FrancescoCioria: not event `t.Object`?

src/tablo/Cell/Cell.js
```diff
@@ -7,6 +7,8 @@ const { maybe, enums, union } = t;
...
+  rowIndex: maybe(t.Integer),
```
FrancescoCioria: why `maybe`? is this necessary for the header row?

src/tablo/Column/Column.js
```diff
@@ -41,7 +41,9 @@ const Column = (args) => {
 
   const cell = ({ rowIndex, columnKey }) => {
     const elem = find([].concat(children), child => child.type === Cell) || defaultCell;
-    return React.cloneElement(elem, { data: (data[rowIndex] || {})[columnKey] });
+    const rowData = data[rowIndex] || {};
+    const dataCell = rowData[columnKey];
+    return React.cloneElement(elem, { data: dataCell, rowData, rowIndex });
```
giogonzo: why `rowIndex` and not e.g. `columnIndex` or `data` (the entire table data)?

src/tablo/Cell/Cell.js
francescogior: The problem is that `Cell` is created before without the props, then `Column` clones it and passes other props to it

src/tablo/Cell/Cell.js
francescogior: It may be an array or anything (Tablo can be used to render a list).
But for the same reason I mentioned below, `t.Any` covers also `undefined`

src/tablo/Column/Column.js
francescogior: Not a strong opinion about `tableData`/`allData` but I just thought that with `rowIndex` one could easily overcome the needs (not typical) to have all the data as argument.

The `columnIndex`/`columnKey` is defined by the user in the `Column.name`

```js
const col = <Column name='columnName'><Cell>{renderData}</Cell></Column>
function renderData(cellData, ...args) { return cellData; }// cellData is `data[rowIndex]['columnName']
```

francescogior: while the `rowIndex` is totally unknown while writing the renderer functions

PR #773, opened 2017-02-16
#770: Tablo: redefine variables for borders (closes #770)

### LINKED ISSUE #770: Tablo: redefine variables for borders
## requirements
Borders widths and styles should be more granularly customisable.
Different variables (with default that makes sense) should be defined for:
- table borders
- cell borders
- resizer line
- header bottom borders

## specs
```
                    ---- table top border ---------------------------------
                     |      | <- fixed-columns-separator   | <- resizer-line   | <- header-cell-border |
                     | ---- header row border ---------                                                    | <-- table right border
table left border  ->|      | <- cell border      |          |   |
                    ---- row border -------------
                    |      |            |          | 
                    ----table bottom border ------
```

img_20170224_122113

For each of the entities, there should be a variable for `width` and one for `color` (used as suffixes)

- `row-border` (default width: 0, default color: cell-border)
- `fixed-columns-separator` (default width is `header-cell-border * 2`, default color is `header-cell-border-color`)
- `cell-border` (default is 

border, width and color)
- `columns-resizer-line` (default is header-cell-border (width and color))
- `header-cell-border`  (the vertical border between header cells) (default is cell-border, width and color)
- `header-row-border` (the horizontal border between header and table body) (default is table border, width and color)
additionally:
- `table-border` (default width and color for the all 4 sides)

francescogior:
Issue #770

### tests performed
- tested on internal-project adding/redefining variables
- showroom works

                              
- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/tablo/Cell/cell.scss
+++ b/src/tablo/Cell/cell.scss
@@ -3,10 +3,14 @@
 
 .tablo {
   .public_fixedDataTableCell_main {
-    border-color: $cell-border-color;
+    border-width: 0; // to reset
+
+    &:not(:last-child) {
+      border-right: $cell-border-width solid $cell-border-color;
+    }
 
     &:last-child {
-      border-width: 0;
+      border-right-width: 0;
     }
 
     .public_fixedDataTableCell_wrap3 {
--- a/src/tablo/Cell/cell.scss
+++ b/src/tablo/Cell/cell.scss
@@ -2,26 +2,32 @@
 @import '../tabloVariables.scss';
 
 .tablo {
-  .public_fixedDataTableCell_main {
-    border-width: 0; // to reset
-
-    &:not(:last-child) {
-      border-right: $cell-border-width solid $cell-border-color;
-    }
-
-    &:last-child {
-      border-right-width: 0;
-    }
-
-    .public_fixedDataTableCell_wrap3 {
-      vertical-align: baseline; // needed to enable `height: 100%` on our cell components
-    }
-  }
-
   .tablo-cell {
     .content {
       padding: 0 5px;
       white-space: nowrap;
+    }
+  }
+
+  .public_fixedDataTableCell_main {
+    &:not(:last-child) {
+      .tablo-cell {
+        border-right: $cell-border-width solid $cell-border-color;
+      }
+    }
+
+    &:last-child {
+      .tablo-cell {
+        border-right: $last-cell-border-width solid $last-cell-border-color;
+
+        &.tablo-cell-fixed {
+          border-right-width: 0;
+        }
+      }
+    }
+
+    .public_fixedDataTableCell_wrap3 {
+      vertical-align: baseline; // needed to enable `height: 100%` on our cell components
     }
   }
 }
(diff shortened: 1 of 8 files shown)


src/tablo/Cell/cell.scss
```diff
@@ -3,10 +3,14 @@
...
-    border-color: $cell-border-color;
+    border-width: 0; // to reset
+
+    &:not(:last-child) {
+      border-right: $cell-border-width solid $cell-border-color;
+    }
 
     &:last-child {
-      border-width: 0;
+      border-right-width: 0;
```
veej: This should be quite useless. Up to you if keep it or not.

src/tablo/Header/header.scss
```diff
@@ -5,26 +5,31 @@
...
+      .tablo-header {
+        color: $header-color;
+        padding: 0 5px;
 
-  .tablo-header {
-    background: $header-background;
-    color: $header-color;
-    padding: 0 5px;
+        /*
+          > div, > div > div {
```
veej: did you investigate on this? could be removed?

src/tablo/Header/header.scss
```diff
@@ -5,26 +5,31 @@
...
-        http://stackoverflow.com/a/25386261;  //TODO(frag) investigate this stuff
-        width: 100%;
+      &:not(:last-child) .tablo-header {
+        border-width: 0; // to reset
+        border-right: $header-cell-border-width solid $header-cell-border-color;
       }
-    */
+
+      &:last-child .tablo-header {
+        border-right-width: 0;
```
veej: This is useless

src/tablo/tabloVariables.scss
```diff
@@ -3,12 +3,38 @@
 $border-color: $brc-silver !default;
 $border-width: 1px !default;
 
+$table-border-color: $border-color !default;
+$table-border-width: $border-width !default;
+$table-border-left-color: $table-border-color !default;
```
veej: Do we really need to have a different color and width for each border here?

src/tablo/Cell/cell.scss
francescogior: This diff should be outdated.
Last column right borders (thus cell, and header) should be default width 0 but customizable because in some projects the content of the table is smaller than the table and you may want to have borders

src/tablo/Header/header.scss
francescogior: Also this is outdated. I removed it in the end. It was commented out since long time and we haven't had issues about vertical centering of the header's content on IE

src/tablo/Header/header.scss
francescogior: See above, it has a purpose (and it's outdated anyway, it changed a little in last revision).

Or maybe I didn't get something?

src/tablo/tabloVariables.scss
francescogior: There are cases where you don't want the bottom border or the right border.
Of course it's hard to find a case when one would use all those variables, but with the default it should work well with minimal setup

veej: Well, I imagined it could be useful to remove some border, but I find very hard we want to have different colors for each of them. Maybe we could keep only 1 var for color and 4 vars for widths?

francescogior: It's basically already the case: `table-border-color`

PR #779, opened 2017-02-21
#727: Popover: debounced events are not canceled when component unmounts (closes #727)

### LINKED ISSUE #727: Popover: debounced events are not canceled when component unmounts
## description
When user pass a value for `delay`, `Popover` debounces `onMouseEnter/Leave` events.
If `Popover` is unmounted before one or more debounced events are fired, `_onMouseEvent` is called when the component is no longer rendered and this throws an error:

image

## how to reproduce
- pass `delay={{ whenClosed: 5000 }}`
- hide `Popover` on `onMouseEnter`
- after 5 seconds `Popover` "wakes-up" and throws the error as it's not rendered anymore

## specs
call `this. onMouseEventDebouncedWhenOpen.cancel()` and `onMouseEventDebouncedWhenClosed.cancel()` in `componentWillUnmount`

## misc
{optional: other useful info}

sawsanias:
Issue #727

### tests performed
* Go to popover > examples > hover.js
* Modify the example as follow:
```javascript
state = {
    renderPopover: true
  };
componentDidMount() {
    setTimeout(() => {
      this.setState({ renderPopover: false });
    }, 4000);
  }
```
* pass `delay={{ whenClosed: 5000 }}`, make sure the delaty of whenClosed > timeout set above
* Add the following:
```javascript
  return this.state.renderPopover ? (
      <Popover popover={popoverProps}>
        <FlexView hAlignContent='center' style={{ width: 150, border: '1px solid #dedede', padding: 10 }}>Hover me!</FlexView>
      </Popover>
    ) : <div>nope</div>;
```

* Run the application `npm run showroom`
* make sure the developertool is opened to see all the errors
* go to the popover component, and check if the error in the issue #727 is shown.

                              
- [x] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/popover/Popover.js
+++ b/src/popover/Popover.js
@@ -76,8 +76,12 @@ export default class Popover extends React.Component {
   componentWillUnmount() {
     this.removePopover();
     this.removeListeners();
-    this.onMouseEventDebouncedWhenOpen.cancel();
-    this.onMouseEventDebouncedWhenClosed.cancel();
+    if (this.onMouseEventDebouncedWhenOpen) {
+      this.onMouseEventDebouncedWhenOpen.cancel();
+    }
+    if (this.onMouseEventDebouncedWhenClosed) {
+      this.onMouseEventDebouncedWhenClosed.cancel();
+    }
 
   }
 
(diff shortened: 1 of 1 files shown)


src/popover/Popover.js
```diff
@@ -76,6 +76,9 @@ export default class Popover extends React.Component {
   componentWillUnmount() {
     this.removePopover();
     this.removeListeners();
+    this.onMouseEventDebouncedWhenOpen.cancel();
```
FrancescoCioria: @sawsanias looking at the code I noticed that `this.onMouseEventDebouncedWhenOpen/Closed` are dynamically defined [here](https://github.com/buildo/react-components/blob/a7acbe0256791373d2e1ca270424e3997922523d/src/popover/Popover.js#L300).

what happens if they're `undefined`? don't these functions end up in a new uncaught `TypeError`?

**EDIT:**
I tried and it doesn't throw any error, this means that we're instantiating `this.onMouseEventDebouncedWhenOpen/Closed` even when we don't need them.

@sawsanias maybe we could fix it directly in this PR?

I'm talking about [this](https://github.com/buildo/react-components/blob/a7acbe0256791373d2e1ca270424e3997922523d/src/popover/Popover.js#L300): if `delayWhenOpen/Closed` is missing `this.onMouseEventDebouncedWhenOpen/Closed` shouldn't be defined.

After this fix I would also call `this.onMouseEventDebouncedWhenOpen/Closed.cancel()` only if they're not `undefined`

PR #785, opened 2017-02-23
#784: Be compatible with `consistent-return` eslint rule  (closes #784)

### LINKED ISSUE #784: Be compatible with `consistent-return` eslint rule
## requirements
- in order to add `consistent-return` to our `eslint-config`without introducing errors we need to change our code in advantage

## specs
- add `consistent-return: 2`  to `.eslintrc` rules
- `npm run lint`
- fix errors (where possible, if possible)

## misc
{optional: other useful info}

ascariandrea:
Closes #784

### tests performed
- run `npm run lint` with `consistent-rule: 2`

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/tablo/plugins/columnsReorder/columnsReorderGrid.js
+++ b/src/tablo/plugins/columnsReorder/columnsReorderGrid.js
@@ -62,8 +62,7 @@ export default (Grid) =>
         const target_index = list.indexOf(target);
         if (source_index <= target_index) {
           return [...list.slice(0, source_index), ...list.slice(source_index + 1, target_index + 1), source, ...list.slice(target_index + 1)];
-        }
-        if (target_index <= source_index) {
+        } else {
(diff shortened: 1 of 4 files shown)


src/tablo/plugins/columnsReorder/columnsReorderGrid.js
```diff
@@ -62,8 +62,7 @@ export default (Grid) =>
         const target_index = list.indexOf(target);
         if (source_index <= target_index) {
           return [...list.slice(0, source_index), ...list.slice(source_index + 1, target_index + 1), source, ...list.slice(target_index + 1)];
-        }
-        if (target_index <= source_index) {
+        } else {
```
FrancescoCioria: this has changed. to reflect previous logic you should keep the `if` and add a `return undefined` at the bottom of the function

/cc @francescogior

FrancescoCioria: 👍

PR #792, opened 2017-02-28
#724: Dropdown: text color should not change on hover (closes #724)

sawsanias:
Closes #724 

**test plan**

- Go to dropDown menu component
- Try to select one of the options in one of the dropDown menu
- Hover over the dropdown, the text color of the **SELECTED VALUE** should **NOT** be changed, as the attached gif:

cksrp6lsoj

PR #793, opened 2017-02-28
#753: Meter: ranges should have a default value (closes #753)

### LINKED ISSUE #753: Meter: ranges should have a default value
## description
- it throws otherwise

## how to reproduce

## specs

## misc
{optional: other useful info}

sawsanias:
Closes #753

### tests performed
* In the examples/default.js, change the passed properties to (remove ranges, and add baseFillingColor)
const meterProps = {
      style: { marginBottom: 20 },
      baseFillingColor: 'brown'
   };

No errors should be shown in the console.

* If a warning message with "You should pass baseFillingColor, ranges are not fully filled"
In the example file of customLabel.js, make sure to add the "baseFillingColor" property to make the warning disappears. 

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/meter/Meter.js
+++ b/src/meter/Meter.js
@@ -48,6 +48,9 @@ export const Props = t.refinement(t.struct({
 }), ({ min, max }) => min < max, 'Props');
 
 const isFullyFilled = (ranges, min, max) => {
+  if (!ranges) {
+    return false;
+  }
   const comparableRanges = ranges.concat({ startValue: max, endValue: min });
   const sortedStartValueList = sortBy(comparableRanges, 'startValue').map(range => range.startValue);
   const sortedEndValueList = sortBy(comparableRanges, 'endValue').map(range => range.endValue);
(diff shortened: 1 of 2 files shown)


FrancescoCioria:
@sawsanias this PR too does not follow our new review process :P
please refer to [this comment](https://github.com/buildo/react-components/pull/792#issuecomment-283103927)

FrancescoCioria:
> using no ranges and passing baseFillingColor and baseBackgroundColor should be considered correct.

@francescogior If I remember correctly the original reqs/specs, `ranges` is indeed optional, but when it exists it should be well tested. By adding a `defaultProp` you're forcing it to be "required" so you can't understand anymore if `ranges` is empty because of a user error or because of the default.
I would personally reconsider removing the `defaultProp` and fixing the logic where it breaks by handling the case `!this.props.ranges` (`ranges` is undefined/null)

sawsanias:
_PS: the warnings were already logged before in the examples. Please fix also them_
It is not clear for me what are the modifications needed on the warning logs. 
If the ranges are not passes, the user should pass both  baseFillingColor and baseBackgroundColor? or one of them is enough?

PR #795, opened 2017-03-01
#794: add the ability to pass custom React context to Modal render tree (closes #794)

### LINKED ISSUE #794: add the ability to pass custom React context to Modal render tree
## requirements
Currently, using `Modal` directly, there's no easy way to render components using arbitrary stuff from context inside the modal.

To be more specific, it might seem to work OK already, because this works:
```js
<Modal>
  <Component foo={this.context.foo} />
</Modal>
```
But this does not:
```js
<Modal>
  <ComponentUsingContextFoo />
</Modal>

// ComponentUsingContextFoo.js :
<View>{this.context.foo}</View>
```

Obviously the reason is that in the first example we are not relying in any way to context to be passed to the new subtree created by `ModalPortal` with `ReactDOM.render()`

Since "context needed throughout the app" is a project-specific thing, there should be a way to pass arbitrary context via `<Modal />` easily

## specs
- receive the context definition via props, and pass it via context to the new render tree created by `ModalPortal`
- since "context providers" typically are initialized passing very top level stuff (not easily available at "leaf" component level), do not require the actual data, just the types, and retrieve it directly via context

e.g.
```js
import { intlShape } from 'react-intl';
// ...
<Modal {...moreProps} childContextTypes={{ intl: intlShape }} />
```
This additional prop would be required only once when customizing the Modal component for the specific project
## misc
{optional: other useful info}

giogonzo:
Closes #794

### tests performed

tested manually on gdsm using
```js
import { modalWithContext } from 'buildo-react-components/lib/modal';
import { intlShape } from 'react-intl';

import './modal.scss';

export default modalWithContext({ intl: intlShape });
```

and adding `@intlMethods` on a component rendered inside the modal

- [x] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/showroom/components.json
+++ b/showroom/components.json
@@ -213,6 +213,11 @@
           {
             "title": "Modal without footer",
             "url": "modal/examples/noFooter.js"
+          },
+          {
+            "title": "Passing arbitrary React context",
+            "description": "Modal renders through a \"portal\", so React context is not propagated as expected.<br><br>*We know context isn't a public api, but it is required for <a href=\"https://github.com/reactjs/react-redux\">widely</a> <a href=\"https://github.com/yahoo/react-intl\">used</a> libraries to work. We also know there's `renderSubtreeIntoContainer`, which is unfortunately even more unstable.*<br/><br/>To solve this problem we export a handy `modalWithContext` that easily allows to pass context down correctly (look at the code!).",
+            "url": "modal/examples/withContext.js"
           }
         ]
       },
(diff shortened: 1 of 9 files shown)


giogonzo:
@FrancescoCioria @veej I was about to add test+example+(props doc?) but if you need this now, go ahead and merge, I'll do in a different PR

giogonzo:
@FrancescoCioria updated with example and a basic test

PR #801, opened 2017-03-08
#800: Dropdown: open the menu on top (closes #800)

### LINKED ISSUE #800: Dropdown: open the menu on top
## requirements
In some cases it could be helpful to have the dropdown menu open at the top instead of bottom (i.e. when dropdown is near the bottom of the page and the menu will overflow the screen)

## specs
- add a new prop `menuPosition` with possible values: `['top', 'bottom']`
- add a new css class when `menuPosition: top`, defining the style as described here: https://github.com/JedWatson/react-select/issues/403

## misc
{optional: other useful info}

veej:
Closes #800

### tests performed

- [x] Google Chrome
- [x] Internet Explorer

### DIFF
--- a/src/dropdown/Dropdown.js
+++ b/src/dropdown/Dropdown.js
@@ -25,6 +25,7 @@ export const Props = {
   flat: t.maybe(t.Boolean),
   autoBlur: t.maybe(t.Boolean),
   simpleValue: t.maybe(t.Boolean),
+  menuPosition: t.enums.of(['top', 'bottom']),
   id: t.maybe(t.String),
   className: t.maybe(t.String),
   style: t.maybe(t.Object)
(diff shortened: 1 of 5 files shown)


src/dropdown/Dropdown.js
```diff
@@ -90,12 +93,13 @@ export default class Dropdown extends React.Component {
...
   getCustomClassNames() {
-    const { size, flat, clearable } = this.props;
+    const { size, flat, clearable, menuPosition } = this.props;
     return cx({
       'is-medium': size === 'medium',
       'is-small': size === 'small',
       'is-flat': flat,
-      'is-clearable': clearable
+      'is-clearable': clearable,
+      top: menuPosition === 'top'
```
francescogior: since the `className` is attached to the top-level component, would it be better to use a more self explaining name? like `has-menu-position-on-top`?

src/dropdown/dropdown.scss
```diff
@@ -194,6 +194,14 @@ $height-small: 30px !default;
     }
   }
 
+  &.top {
```
francescogior: and in case, change it here of course

PR #806, opened 2017-03-09
#798: Dropdown: implement grouped options (closes #798)

### LINKED ISSUE #798: Dropdown: implement grouped options
526c5644-034c-11e7-92da-baaaa032da0d

## requirements
- add the ability to group options under groups

## specs
- use menuRenderer (see the default here: https://github.com/JedWatson/react-select/blob/master/src/utils/defaultMenuRenderer.js), leaving the possibility to override the default by the user
- support `Option` object to accept an optional `optionGroup` string parameter. The parameter can be of any name. 
- `'optionGroup'` should be just the default `groupByKey` prop, overridable by the user
- add `optionGroupRenderer` optional prop (`t.maybe(t.Function))`) to be able to customize the rendering of the group title
- add a `defaultOptionGroupRender` to render the default style (see mockup above)
- options not declaring `optionGroup` should be grouped in a group without title (the current behaviour when none the options declare `optionGroup`). This should be handled by the `defaultGroupRenderer`. Be sure to render the `undefined` group as first
- use

```js
  _.map(_.groupBy(options, this.props.groupByKey), (optionGroup, optionGroupTitle) => (
<OptionGroup>
  {this.props.optionGroupRender({ optionGroupTitle })}
  optionGroup.map(x => (
   <Option ...
....

```

## misc
{optional: other useful info}

sawsanias:
Closes #798

### tests performed
* starting the showroom locally, and check the group example:
image

* Got to the src/dropdown/examples/groups.js
and try to modify some of the categories for testing, for example adding an option without category
`  const options = [
      { value: 'apple', label: 'Apple', category: 'Fruit'},
      { value: 'orange', label: 'Orange', category: 'Fruit' },
      { value: 'mandarin', label: 'Mandarin', category: 'Fruit' },
      { value: 'onion', label: 'Onion', category: 'Vegetable' },
      { value: 'peach', label: 'Peach', category: 'Fruit' },
      { value: 'apricot', label: 'Apricot', category: 'Fruit' },
      { value: 'tomato', label: 'Tomato', category: 'Vegetable' },
      { value: 'pineapple', label: 'Pineapple', category: 'Fruit' },
      { value: 'cucumber', label: 'Cucumber', category: 'Vegetable'},
      { value: 'banana', label: 'Banana'}
    ];`
then see the results, the options with undefined groupByKey shouldn't appear at the top of the list without any group, as:
image

* Make sure all the others examples are still working


src/dropdown/Dropdown.js
```diff
@@ -115,10 +133,85 @@ export default class Dropdown extends React.Component {
...
+    else {
+      return (
+        <FlexView className='Select-option-group'>
+          {this.props.optionGroupRenderer(title)}
+        </FlexView>
+      );
+    }
+  }
+
+  _menuRenderer = ({
```
veej: I'd call this `defaultMenuRenderer` and I'll pass it as defaultProp exactly in the same way you're doing for `optionGroupRenderer`.
Also, you're completely ignoring the `menuRenderer` coming from props.

src/dropdown/Dropdown.js
```diff
@@ -131,7 +224,8 @@ export default class Dropdown extends React.Component {
       className: cx('dropdown', className, this.getCustomClassNames()),
       value: this.valueToOption(this.getValue(), options),
       onInputKeyDown,
-      onChange: _onChange
+      onChange: _onChange,
+      menuRenderer: _menuRenderer
```
veej: As already mentioned, you're completely ignoring the `menuRenderer` possibly coming from props. I'd set `_menuRenderer` as the default value for `menuRenderer` prop.

src/dropdown/dropdown.scss
```diff
@@ -168,6 +170,14 @@ $height-small: 30px !default;
     background-color: $menu-background;
     box-shadow: rgba(34, 36, 38, .12) 0 2px 4px 0;
 
+    .Select-option-group {
+      padding: 0 9px;
+      font-size: 15px;
+      font-weight: 600;
```
veej: please, use `$brc-semibold` here

src/dropdown/Dropdown.js
```diff
@@ -115,10 +133,85 @@ export default class Dropdown extends React.Component {
...
+    optionComponent,
+    optionRenderer,
+    options,
+    valueArray,
+    valueKey
+  }) => {
+
+    const Option = optionComponent;
+    const propertyName = this.props.groupByKey;
+    const sortedOptions = sortBy(options, (option) => {
```
veej: Pay attention here: you're changing the order in which options are rendered, but you're passing them to the underlying `react-select` keeping the original sorting. When you move between options using arrow keys, options are focused in the order you passed them to the `react-select` and not in the order they're rendered:



I suggest to sort them  in `getLocals` and not here.

src/dropdown/Dropdown.js
sawsanias: Actually I used it as onChange property.

PR #812, opened 2017-03-16
#810: Dropdown: in multi mode, values should wrap and input height should adapt size (closes #810)

### LINKED ISSUE #810: Dropdown: in multi mode, values should wrap and input height should adapt size

sawsanias:
Closes #810

### tests performed

* the new multi dropdown:
image

_**Note**: the delete icon is on the whole block, and not only on the icon._
In the screenshot below, the mouse was hover the container of the x
image

- [ ] Google Chrome
- [ ] Internet Explorer


src/dropdown/dropdown.scss
```diff
@@ -213,23 +213,59 @@ $height-small: 30px !default;
...
+      padding: 0;
+    }
+
+    .Select-placeholder {
+      padding: 0 10px;
+    }
+
+    .Select-control {
+      padding: 0;
+      display: table;
```
veej: Changing the `display` value to table is really dangerous, since we're forcing the whole dropdown to use `display: flex` (see the `normalize` mixin defined above) . You should be able to accomplish your task using `@include flex_flex-wrap(wrap)` where needed

src/dropdown/dropdown.scss
```diff
@@ -213,23 +213,59 @@ $height-small: 30px !default;
...
+      vertical-align: middle;
+    }
+
+    .Select-multi-value-wrapper {
+      width: auto;
+    }
+
     .Select-control .Select-value {
       margin-right: 5px;
+      margin-left: 5px;
```
veej: here you're adding an additional `5px` margin between every value. I supposed your intent was to add space from left-size of the dropdown. Maybe adding a padding where needed is a better solution.

src/dropdown/dropdown.scss
```diff
@@ -213,23 +213,59 @@ $height-small: 30px !default;
...
+      margin-top: 5px;
+      vertical-align: top;
       border: 1px solid $multiple-value-border-color;
       background: $multiple-value-background-color;
       color: $multiple-value-color;
-
-      @include height($height - 14px);
```
veej: Why are you changing the height of values?

src/dropdown/dropdown.scss
```diff
@@ -213,23 +213,59 @@ $height-small: 30px !default;
...
+      display: inline-block;
+      height: $height - 12px;
+      line-height: $height - 14px;
 
       .Select-value-label {
-        @include height($height - 14px);
-
         padding-top: 0;
         padding-bottom: 0;
+        width: auto;
```
veej: Is this needed?

src/dropdown/dropdown.scss
```diff
@@ -246,6 +282,31 @@ $height-small: 30px !default;
...
+      padding-right: 5px;
+    }
+
+    .Select-arrow {
+      display: inline-block;
+      position: relative;
+    }
+
+    .Select-clear-zone {
+      width: 17px;
```
veej: Why are you changing this width?

src/dropdown/dropdown.scss
```diff
@@ -213,23 +213,59 @@ $height-small: 30px !default;
   }
 
   &.Select--multi {
+    height: auto;
```
veej: My fault, this is not true... `size` it's working, but there's some UI defect if `size='small'`

francescogior:
closing in favor of #813

PR #815, opened 2017-03-20
#814: dropdown: sass variables for `.Select-value-label` is overridden (closes #814)

### LINKED ISSUE #814: dropdown: sass variables for `.Select-value-label` is overridden
## description
Setting this variable in a project has no effect

## how to reproduce

## specs
- add `.has-value` to the relevant selector to make it beat the declaration of the default color 

## misc
{optional: other useful info}

francescogior:
Closes #814

### tests performed
image


src/dropdown/dropdown.scss
```diff
@@ -236,18 +236,18 @@ $height-small: 30px !default;
       position: relative;
     }
 
-    .Select-control .Select-value {
+    &.has-value .Select-control .Select-value {
```
ascariandrea: You're changing the CSS path here.
No regressions are caused, right?

francescogior: `.Select-value` exists only if the `.dropdown` has `.has-value`, then I'm sure no regression will be introduced by this change

ascariandrea: Then ok!

PR #818, opened 2017-03-22
#817: AsyncStatusIndicator should export `SavingState` model (closes #817)

### LINKED ISSUE #817: AsyncStatusIndicator should export `SavingState` model
## requirements
as per title

## specs
The module itself is already exporting its `Props` but those are not exported from the index, thus they are unreachable (i.e. `import { Props } from 'buildo-react-components/lib/async-status-indicator'` doesn't work, while it should. Still `import { Props } from 'buildo-react-components/lib/async-status-indicator/AsyncStatusIndicator' works).
Also, `import { AsyncStatusIndicatorState } from 'buildo-react-components/lib/async-status-indicator'` should work

- export `Props as AsyncStatusIndicatorProps` from `async-status-indicator/index`
- export `AsyncStatusIndicatorState` from the module and from the index

## misc
{optional: other useful info}

sawsanias:
Closes #817

### tests performed
* install locally the react-components to  alinityPro project to test. 
* In alinityPro project use the exported AsyncStatusIndicatorState in: 
src/app/components/Basic/AsyncStatusIndicator/AsyncStatusIndicator.js
export { AsyncStatusIndicatorState as SavingState } from 'buildo-react-components/lib/async-status-indicator';
* make sure the async status is still working.

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/async-status-indicator/index.js
+++ b/src/async-status-indicator/index.js
@@ -1,2 +1,4 @@
-import AsyncStatusIndicator from './AsyncStatusIndicator';
+import AsyncStatusIndicator, { Props as AsyncStatusIndicatorProps, AsyncStatusIndicatorState } from './AsyncStatusIndicator';
+
+export { AsyncStatusIndicatorProps, AsyncStatusIndicatorState };
 export default AsyncStatusIndicator;
--- a/src/async-status-indicator/AsyncStatusIndicator.js
+++ b/src/async-status-indicator/AsyncStatusIndicator.js
@@ -3,7 +3,7 @@ import { pure, skinnable, props, t } from '../utils';
 import cx from 'classnames';
 import View from 'react-flexview';
 
-const AsyncStatusIndicatorState = t.enums.of([
+export const AsyncStatusIndicatorState = t.enums.of([
   'ready',
   'processing',
   'success',


src/async-status-indicator/index.js
```diff
@@ -1,2 +1,4 @@
-import AsyncStatusIndicator from './AsyncStatusIndicator';
-export default AsyncStatusIndicator;
\ No newline at end of file
+import AsyncStatusIndicator, { Props as AsyncStatusIndicatorProps, AsyncStatusIndicatorState } from './AsyncStatusIndicator';
+
+export { AsyncStatusIndicatorProps, AsyncStatusIndicatorState };
+export default AsyncStatusIndicator;
```
FrancescoCioria: NOTE: I'm not 100% sure but I think you can export everything with this one liner:

```js
export default, { Props as AsyncStatusIndicatorProps, AsyncStatusIndicatorState } from './AsyncStatusIndicator';
```

francescogior: I confirm that the one liner is working, but in the end there's no a strict guideline about how to do that. (There are mixed usages in the app)

PR #821, opened 2017-03-22
Greenkeeper/stylelint webpack plugin 0.7.0

FrancescoCioria:
Closes #757


FrancescoCioria:
Apparently we've been relying on a defect (fixed in `0.6.0`?) that caused webpack to fail even with `failOnError: false` :P
I added a commit to this PR to fix it

PR #826, opened 2017-03-23
#825: timePicker doesn't update the options with new input value (closes #825)

### LINKED ISSUE #825: timePicker doesn't update the options with new input value
## description
When the user manually enter an input value the did not already exist in the options, `TimePicker` doesn't update the options list with the new value.

## how to reproduce
* in one of the examples of the timepicker component initialize a state value of non existed option, for example:
state = { value: { hours: 10, minutes: 32 }}
* start the application: the timepicker of the example would not populate the initial value.

## specs
In the function of makeOptions, make sure to add the inputValue.

## misc
{optional: other useful info}

sawsanias:
Closes #825

## Test Plan

* run the application, and check the 24H TimePicker, the value 11:23 should be selected.
825-timepicker_doesnt_update_the
image

* Open the dropdown menu, and see the option 11:23 within the options and in the right order
image

* retry by entering new value e.g. 08:33
image

* retry by enter the value 08:02 (value is not within the range)
image

* Change the state value of TimePicker12.js by: `state = { value: { hours: 11, minutes: 23 } }`
image
 start typing 11:
image

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -141,13 +141,12 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
 
 export const makeOptions = ({ minTime, maxTime, timeFormat, userValue }, inputValue) => {
   const time = parseInTimeFormat(inputValue, timeFormat);
-  //make sure the userValue has timeFormat so in H12 the day period will be shown
   if ( userValue && userValue !== inputError ) {
     userValue.timeFormat = timeFormat;
   }
-  const timeList = time === inputError ? [] : createTimeList(time, timeFormat).concat(compact([userValue]));
+  const timeList = (time === inputError ? [] : createTimeList(time, timeFormat)).concat(compact([userValue]));
   const filteredTimeList = timeList.filter(filterTime({
-    originalInput: time.originalInput, minTime, maxTime
+    originalInput: time.originalInput || '', minTime, maxTime
   }));
   return uniqBy(sortBy(filteredTimeList.map(toOption), 'value'), 'value');
 };
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -141,10 +141,8 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
 
 export const makeOptions = ({ minTime, maxTime, timeFormat, userValue }, inputValue) => {
   const time = parseInTimeFormat(inputValue, timeFormat);
-  if ( userValue && userValue !== inputError ) {
-    userValue.timeFormat = timeFormat;
-  }
-  const timeList = (time === inputError ? [] : createTimeList(time, timeFormat)).concat(compact([userValue]));
+  const selectedValue = (userValue && userValue !== inputError) ? { ...userValue, timeFormat } : userValue;
+  const timeList = (time === inputError ? [] : createTimeList(time, timeFormat)).concat(compact([selectedValue]));
   const filteredTimeList = timeList.filter(filterTime({
     originalInput: time.originalInput || '', minTime, maxTime
   }));
--- a/src/time-picker/TimePicker.js
+++ b/src/time-picker/TimePicker.js
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import Dropdown from '../dropdown/Dropdown';
 import range from 'lodash/range';
 import flatten from 'lodash/flatten';
+import compact from 'lodash/compact';
 import uniqBy from 'lodash/uniqBy';
 import sortBy from 'lodash/sortBy';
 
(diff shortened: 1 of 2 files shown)


sawsanias:
@francescogior I am not sure about H12 timeformat, should I consider to populate just one value?

src/time-picker/TimePicker.js
```diff
@@ -136,13 +139,17 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
...
 };
 
-export const makeOptions = ({ minTime, maxTime, timeFormat }, inputValue) => {
+export const makeOptions = ({ minTime, maxTime, timeFormat, userValue }, inputValue) => {
   const time = parseInTimeFormat(inputValue, timeFormat);
-  const timeList = time === inputError ? [] : createTimeList(time, timeFormat);
+  //make sure the userValue has timeFormat so in H12 the day period will be shown
+  if ( userValue && userValue !== inputError ) {
+    userValue.timeFormat = timeFormat;
+  }
```
francescogior: remove 144:147, you're not using `timeFormat` anywhere

src/time-picker/TimePicker.js
```diff
@@ -136,13 +139,17 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
...
+  const timeList = time === inputError ? [] : createTimeList(time, timeFormat).concat(compact([userValue]));
```
francescogior: The correct way would be
```js
const timeList = (time === inputError ? [] : createTimeList(time, timeFormat)).concat(compact([userValue]));
```
ie you concat the `userValue` even if the `time === `inputError` (but see related comment below)

src/time-picker/TimePicker.js
```diff
@@ -136,13 +139,17 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
...
   const filteredTimeList = timeList.filter(filterTime({
     originalInput: time.originalInput, minTime, maxTime
```
francescogior: `originalInput: time.originalInput || ''` because it can be undefined and it'd break in the `filterTime` (precisely in the `startsWith`) function

francescogior:
@sawsanias also there's a eslint error to fix that breaks the ci

`144:3  error  Expected space or tab after '//' in comment  spaced-comment`

src/time-picker/TimePicker.js
sawsanias: if you don't set the TimeFormat, it will consider it as H24 and will show the option as:

image


it is used in 

`const formatTime = ({ hours, minutes, timeFormat }) => (
  timeFormat === H12 ? formatTime12({ hours, minutes }) : formatTime24({ hours, minutes })
);`

francescogior: You're right!
Still, I don't like to modify function parameters. I'll leave a comment in the next review

src/time-picker/TimePicker.js
```diff
@@ -141,13 +141,12 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
 
 export const makeOptions = ({ minTime, maxTime, timeFormat, userValue }, inputValue) => {
   const time = parseInTimeFormat(inputValue, timeFormat);
-  //make sure the userValue has timeFormat so in H12 the day period will be shown
   if ( userValue && userValue !== inputError ) {
```
francescogior: do not update function parameters:
```
const selectedValue = (userValue && userValue !== inputError) ? { ...userValue, timeFormat } : userValue
```

src/time-picker/TimePicker.js
```diff
@@ -141,13 +141,12 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
...
     userValue.timeFormat = timeFormat;
   }
-  const timeList = time === inputError ? [] : createTimeList(time, timeFormat).concat(compact([userValue]));
+  const timeList = (time === inputError ? [] : createTimeList(time, timeFormat)).concat(compact([userValue]));
```
francescogior: and then 
```js
...)).concat(compact([selectedValue]);
```

src/time-picker/TimePicker.js
```diff
@@ -136,13 +139,14 @@ export const filterTime = ({ originalInput, minTime, maxTime }) => time => {
...
 
-export const makeOptions = ({ minTime, maxTime, timeFormat }, inputValue) => {
+export const makeOptions = ({ minTime, maxTime, timeFormat, userValue }, inputValue) => {
   const time = parseInTimeFormat(inputValue, timeFormat);
-  const timeList = time === inputError ? [] : createTimeList(time, timeFormat);
+  const selectedValue = (userValue && userValue !== inputError) ? { ...userValue, timeFormat } : userValue;
+  const timeList = (time === inputError ? [] : createTimeList(time, timeFormat)).concat(compact([selectedValue]));
   const filteredTimeList = timeList.filter(filterTime({
-    originalInput: time.originalInput, minTime, maxTime
+    originalInput: time.originalInput || '', minTime, maxTime
```
ascariandrea: I would edit this function (or add a second `.filter` step) by filtering also `undefined` values to avoid the `.concat(compact([selectedValue]))`.

francescogior:
tested on internal project and fix the related bug 👍

PR #829, opened 2017-03-27
#827: dropdown:`Select-multi-value-wrapper` should be vertically centered (closes #827)

### LINKED ISSUE #827: dropdown:`Select-multi-value-wrapper` should be vertically centered
## description
image

## how to reproduce

## specs
{optional: describe a possible fix for this defect, if not obvious}

## misc
{optional: other useful info}

FrancescoCioria:
Closes #827

### tests performed

multi values centered:
image

multi values centered and well spaced when wrapping:
image

size "small":
image

- [x] Google Chrome
- [x] Internet Explorer

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -238,8 +238,8 @@ $height-small: 30px !default;
 
     &.has-value .Select-control .Select-value {
       margin-right: 5px;
-      margin-top: auto;
-      margin-bottom: auto;
+      margin-top: 5px;
+      margin-bottom: 5px;
       border: 1px solid $multiple-value-border-color;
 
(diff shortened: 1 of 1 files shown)


FrancescoCioria:
> Is possible to reduce the space between the rows?

yes but then the last row would not be centered properly because of the input (`.Select-input`) which is 36px height (as I've written in the comment the 10px vertical margin is not random but chosen to fill this gap as the value is 26px tall):

image

We could make the input 26px tall but it will probably cause new defects as usual :P

francescogior:
> We could make the input 26px tall but it will probably cause new defects as usual 

Ok, but if there's other way I think we should at least try :)

FrancescoCioria:
image

PR #837, opened 2017-03-30
#830: Dropdown: on IE11, values in multi-select mode are not vertically aligned when in single line (closes #830)

### LINKED ISSUE #830: Dropdown: on IE11, values in multi-select mode are not vertically aligned when in single line
## description
On IE11, values in multi-select mode are not vertically aligned **when in single line**

image

## how to reproduce

## specs

## misc
{optional: other useful info}

sawsanias:
Closes #830

### tests performed
* In IE11, test the default size dropdown:
image
* Add more options till they wrap to next line:
image

*Change the size of the dropdown in the multi example by adding the property `size='small'`, in IE11
image

*Continue adding options till they wrapped to  new line:
image

* Add the property searchable to dropdown and test the dropdown
upvphrcmj1

Repeat the steps above in Chrome, and make sure everything is working.

- [x] Google Chrome
- [x] Internet Explorer

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -221,7 +221,14 @@ $height-small: 30px !default;
       height: auto;
 
       .Select-multi-value-wrapper {
+        min-height: $height;
         height: auto;
+      }
+    }
+
+    &.is-small {
+      .Select-multi-value-wrapper {
+        min-height: $height-small;
       }
     }
 
@@ -229,6 +236,7 @@ $height-small: 30px !default;
       .Select-control {
         .Select-multi-value-wrapper {
           padding: 2px 0;
+          min-height: auto;
 
           .Select-input {
             position: relative;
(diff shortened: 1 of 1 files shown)


src/dropdown/dropdown.scss
```diff
@@ -221,17 +221,10 @@ $height-small: 30px !default;
       height: auto;
 
       .Select-multi-value-wrapper {
-        min-height: $height;
         height: auto;
```
francescogior: removing this makes disappear the placeholder

francescogior: And the input: if you make it `searchable` you don't see what you type (if there are no value)

I'm not 100% sure, but I think you can fix it using the `.has-value` class and apply the min-height conditionally

src/dropdown/dropdown.scss
```diff
@@ -287,6 +288,8 @@ $height-small: 30px !default;
 
       &.is-small {
         .Select-control .Select-multi-value-wrapper {
+          min-height: auto;
```
veej: This looks useless... the previous line seems to include this case too

francescogior: I think it is to beat the `.is-small .Select-multi-value-wrapper { min-height: $height-small; }`

PR #851, opened 2017-04-05
#850: Dropdown: no style for focus (closes #850)

### LINKED ISSUE #850: Dropdown: no style for focus
## description
Border-color should change when the dropdown is focused

## how to reproduce
- focus on a dropdown

## specs
- Add a new variable: `$border-color-focused` (with default value `$waterBlue`) and use it when needed

## misc
{optional: other useful info}

veej:
Closes #850

### tests performed
- when focused hitting TAB

- when disabled, border doesn't change on focus

- [x] Google Chrome
- [x] Internet Explorer

### DIFF
--- a/src/dropdown/dropdown.scss
+++ b/src/dropdown/dropdown.scss
@@ -6,6 +6,7 @@ $background-hover: $brc-gradientDarkGrey !default;
 $background-open: $brc-gradientDarkGrey !default;
 $border-color: $brc-silver !default;
 $border-color-open: $brc-waterBlue !default;
+$border-color-focus: $brc-waterBlue !default;
 $border-radius: 4px !default;
 $text-color: $brc-coolGrey !default;
 $text-color-open: $brc-coolGrey !default;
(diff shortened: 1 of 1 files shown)


FrancescoCioria:
@veej have you verified that when the dropdown is open `$border-color-open` beats `$border-color-focus`?

francescogior:
@FrancescoCioria it does
image

FrancescoCioria:
> this should be considered breaking, what do you think?

If I'm not wrong, In the original paper written by alessandra the focus was there which means that this is a simple defect.
Nevertheless, as we've kept this defect for maaaaaaany months it may be one of those cases where it's better to upgrade it to "breaking".

veej:
> @veej have you verified that when the dropdown is open $border-color-open beats $border-color-focus?

@FrancescoCioria yep

> this should be considered breaking, what do you think?

@francescogior same old doubt. This could be considered a defect (`react-select` has a style for focus), but yeah, for anyone who doesn't want a style for focus this is potentially breaking.
Btw I won't use `$border-color` as default value, since default values should follow our guidelines.

francescogior:
> Btw I won't use $border-color as default value, since default values should follow our guidelines.

Right. Then breaking and devs should take care to backport this on relevant projects

PR #869, opened 2017-04-14
#737: Implement FormattedParagraph (closes #737)

### LINKED ISSUE #737: Implement FormattedParagraph
## requirements
formatting a string like `Hi Ambrogio,\n\nI'm writing you...` is very annoying as HTML collapses groups of `\n` in a single one.

As it's a very common situation, it would be useful to have component (we already have one in oxway and IperCron) to format `\n` correctly.

This component should also accept a nice `paragraphSpacing` prop to let easily define between paragraphs.

A new paragraph should start after a series of `\n` >= 2.
A single `\n` should be mapped to a simple `<br />`

The `marginTop` of each paragraph should be proportional to the length of the series of `\n`:
- 2 -> `marginTop: paragraphSpacing`
- 3 -> `marginTop: 2 * paragraphSpacing`
- ...

## specs
start from `FormattedParagraph` component from IperCron

## misc
{optional: other useful info}

FrancescoCioria:
Closes #737

### tests performed
image

- [x] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/formatted-paragraph/FormattedParagraph.js
+++ b/src/formatted-paragraph/FormattedParagraph.js
@@ -35,13 +35,15 @@ export default class FormattedParagraph extends React.Component {
   }
 
   getLocals({ content, paragraphSpacing, ...props }) {
+    const { Paragraph } = this;
+
     const seriesOfNewParagraphs = content.match(/\n\n+/g);
     const paragraphs = content.split(/\n\n+/);
 
     const children = paragraphs.map((paragraph, i) => (
-      <this.Paragraph marginTop={i !== 0 ? paragraphSpacing * (seriesOfNewParagraphs[i - 1].length - 1) : 0}>
+      <Paragraph marginTop={i !== 0 ? paragraphSpacing * (seriesOfNewParagraphs[i - 1].length - 1) : 0}>
         {paragraph}
-      </this.Paragraph>
+      </Paragraph>
     ));
 
     return {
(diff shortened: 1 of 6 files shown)


src/formatted-paragraph/FormattedParagraph.js
```diff
@@ -0,0 +1,60 @@
...
+
+    return (
+      <p style={style}>
+        {_children}
+      </p>
+    );
+  }
+
+  getLocals({ content, paragraphSpacing, ...props }) {
+    const seriesOfNewParagraphs = uniq(content.match(/\n\n+/g));
```
veej: `uniq` here seems to be wrong. If you remove paragraphs here, `seriesOfNewParagraph[i - 1].length` will fail.

Try using `a\n\nb\n\nc` as content.

src/formatted-paragraph/FormattedParagraph.js
```diff
@@ -0,0 +1,60 @@
...
+    const paragraphs = content.split(/\n\n+/);
+
+    const children = paragraphs.map((paragraph, i) => (
+      <this.Paragraph marginTop={i !== 0 ? paragraphSpacing * (seriesOfNewParagraphs[i - 1].length - 1) : 0}>
```
veej: I can't really stand `<this.` 😖
Usually I define `const { Paragraph } = this;`

src/formatted-paragraph/FormattedParagraph.js
FrancescoCioria: I thought I had already removed it!!! thanks 👍

src/formatted-paragraph/FormattedParagraph.js
FrancescoCioria: ooook 👨‍🎨 :)

PR #874, opened 2017-04-18
#873: replace FormattedParagraph with FormattedText (closes #873)

### LINKED ISSUE #873: replace FormattedParagraph with FormattedText
## requirements
FormattedText should:
- replace `\n` with `<br>` (simpler logic!)
- return a `span`
- replace links with `<a href={link} target='_blank'>{originalText}</a>`

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Closes #873

### tests performed
The example renders into this:
image

- it doesn't crash if `children` has no `\n`
- it doesn't crash if `children` has no link

- [x] Google Chrome
- [ ] Internet Explorer


src/formatted-text/FormattedText.js
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import { props, t, skinnable } from '../utils';
+import flattenDeep from 'lodash/flattenDeep';
+import tlds from 'tlds';
+
+const linkify = require('linkify-it')().tlds(tlds);
+
+export const Props = {
+  content: t.String,
```
giogonzo: wouldn't this be better as `children`?

src/formatted-text/FormattedText.js
```diff
@@ -0,0 +1,83 @@
...
+    if (matches) {
+      return matches.reduce((acc, match, i) => {
+        const nextMatch = matches[i + 1];
+
+        return [
+          ...(acc || [string.slice(0, match.index)]),
+          <a href={match.url} target='_blank'>{match.raw}</a>,
+          string.slice(match.lastIndex, nextMatch ? nextMatch.index : undefined)
+        ];
+      }, null);
```
giogonzo: I'm having quite an hard time reading this `reduce`, maybe `matches.reduce((s, match) => s.replace(match.raw, 'LINK'), initialString)` would be simpler?
but not sure, feel free to ignore

giogonzo: discussed offline, it doesn't make much sense since we'd need to set `innerHTML` with this approach

PR #945, opened 2017-06-27
#917: [typescript .d.ts] Badge (closes #917)

### LINKED ISSUE #917: [typescript .d.ts] Badge
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #917

### tests performed
- tested on `alinity-pro` on `ts-brc` branch ✅ 

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/badge/Badge.d.ts
+++ b/src/badge/Badge.d.ts
@@ -1,0 +1,10 @@
+import { PureComponent } from 'react';
+
+export type BadgeProps = {
+  label: any,
+  active?: boolean,
+  className?: string,
+  style?: object
+}
+
+export default class Badge extends PureComponent<BadgeProps, void> {}
--- a/src/badge/index.d.ts
+++ b/src/badge/index.d.ts
@@ -1,0 +1,3 @@
+import Badge, { BadgeProps } from './Badge';
+export default Badge;
+export { BadgeProps };


src/Divider/Divider.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare namespace Divider {
+
+  export type Orientation = 'horizontal' | 'vertical';
+  export type Size = 'small' | 'medium' | 'large' | 'no-margin';
+
+  type Props = {
+    orientation: Orientation,
+    size: Size,
+    style: object
```
FrancescoCioria: all these props should be optional (append a `?` to each key)

src/Divider/Divider.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+declare class Divider extends PureComponent<Divider.Props, void> {}
+
+export = Divider
```
FrancescoCioria: mmm why isn't this an `export default` like the `js` file?

src/badge/Badge.d.ts
```diff
@@ -0,0 +1,14 @@
+import { PureComponent, ReactChildren } from 'react';
+
+declare namespace Badge {
+
+  type Props = {
+    label: ReactChildren,
```
FrancescoCioria: Typescript's `ReactChildren` is quite different from `tcomb`'s one
I'm pretty sure this is not enough...

I tried this quickly in a project and it complains:
```ts
type X = {
    child: React.ReactChildren
  };

    const x: X = {
      child: <div />
    };

    const x: X = {
      child: ''
    };
```

src/badge/Badge.js
```diff
@@ -35,4 +34,8 @@ export default class Badge extends React.PureComponent {
       </FlexView>
     );
   }
+
+  render() {
+    return this.template(this.getLocals());
```
FrancescoCioria: why this change? we're shouldn't touch `.js` files in these issues

src/button/Button.d.ts
```diff
@@ -0,0 +1,35 @@
...
+    type?: ButtonType,
+    primary?: boolean,
+    flat?: boolean,
+    size?: ButtonSize,
+    fluid?: boolean,
+    circular?: boolean,
+    textOverflow?: React.ComponentClass<object>, // todo(typo) when translate text overflow, update here its props types
+    style?: object,
+    className?: string,
+    baseState?: ButtonBaseState,
```
FrancescoCioria: these last three props should be removed as they are from `StatefulButton`

src/button/StatefulButton.d.ts
```diff
@@ -0,0 +1,10 @@
+import { PureComponent } from 'react';
+import * as Button from './Button';
+
+declare namespace StatefulButton {
+  type Props = Button.Props;
```
FrancescoCioria: and they should be added here with an `&`:
```
type Props = Button.Props & { baseState?: ...}
```

src/text-overflow/TextOverflow.d.ts
```diff
@@ -0,0 +1,16 @@
+import { PureComponent } from 'react';
+
+declare namespace TextOverflow {
+  type Props = {
+    label?: string | number,
```
FrancescoCioria: `label` is required and should accept only `string`

src/text-overflow/TextOverflow.d.ts
```diff
@@ -0,0 +1,16 @@
...
+    popover?: object, // TODO(typo): Omit?
+    lazy?: boolean,
+    id?: string,
+    className?: string,
+    style?: object
+  }
```
FrancescoCioria: you should reflect exactly the `Props` object from `TextOverflow.js`

here you added `popover` while `children` and `delayWhenLazy` are both missing.

src/button/Button.d.ts
```diff
@@ -0,0 +1,35 @@
...
+    label?:string | object,
+    icon?:string | object,
+    children?: string,
+    type?: ButtonType,
+    primary?: boolean,
+    flat?: boolean,
+    size?: ButtonSize,
+    fluid?: boolean,
+    circular?: boolean,
+    textOverflow?: React.ComponentClass<object>, // todo(typo) when translate text overflow, update here its props types
```
FrancescoCioria: you did it in this PR, right? can't you use it here yet?

src/Divider/Divider.d.ts
ascariandrea: cause write as this let us to:
```ts
import * as Divider from 'path/to/divider';

class MyDivider extender React.PureComponent<Divider.Props, void> {
 render() {
  return <Divider {...this.props} />
 }
}
```

PR #949, opened 2017-06-28
#937: [typescript .d.ts] TextOverflow (closes #937)

### LINKED ISSUE #937: [typescript .d.ts] TextOverflow
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #937

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/text-overflow/TextOverflow.d.ts
+++ b/src/text-overflow/TextOverflow.d.ts
@@ -1,0 +1,12 @@
+import { PureComponent } from 'react';
+
+export type TextOverflowProps = {
+  label?: string | number,
+  popover?: object, // TODO(typo): Omit?
+  lazy?: boolean,
+  id?: string,
+  className?: string,
+  style?: object
+}
+
+export default class TextOverflow extends PureComponent<TextOverflowProps, void> {}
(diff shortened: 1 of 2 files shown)


src/text-overflow/TextOverflow.d.ts
```diff
@@ -0,0 +1,12 @@
+import { PureComponent } from 'react';
+
+export type TextOverflowProps = {
+  label?: string | number,
+  popover?: object, // TODO(typo): Omit?
+  lazy?: boolean,
+  id?: string,
+  className?: string,
+  style?: object
```
FrancescoCioria: same comments as before: `TextOverflowProps` should reflect exactly the `@props` from `TextOverflow.js`.

- `label` is required and only string,
- `children` is missing
- `delayWhenLazy` is missing
- `popover` is not a prop of TextOverflow

Also, `style` should be `React.CSSProperties` because `object` represents any non-primitive type (which means also arrays, regexp, dates...)

ascariandrea: Fixed!

PR #950, opened 2017-06-28
#919: [typescript .d.ts] Button (closes #919)

### LINKED ISSUE #919: [typescript .d.ts] Button
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #919

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/button/Button.d.ts
+++ b/src/button/Button.d.ts
@@ -1,0 +1,29 @@
+import * as React from 'react';
+
+export namespace ButtonProps {
+
+  type ButtonState = 'ready' | 'not-allowed' | 'processing' | 'error' | 'success';
+  type ButtonType = 'default' | 'primary' | 'positive' | 'negative' | 'flat';
+  type ButtonSize = 'tiny' | 'small' | 'medium';
+
+}
+
+export type ButtonProps = {
+    onClick: (e: React.SyntheticEvent<HTMLDivElement>) => void,
+    buttonState?: ButtonProps.ButtonState,
+    label?:string | object,
+    icon?:string | object,
+    children?: string,
+    type?: ButtonProps.ButtonType,
+    primary?: boolean,
+    flat?: boolean,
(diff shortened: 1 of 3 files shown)


src/button/Button.d.ts
```diff
@@ -0,0 +1,29 @@
...
+  type ButtonState = 'ready' | 'not-allowed' | 'processing' | 'error' | 'success';
+  type ButtonType = 'default' | 'primary' | 'positive' | 'negative' | 'flat';
+  type ButtonSize = 'tiny' | 'small' | 'medium';
+
+}
+
+export type ButtonProps = {
+    onClick: (e: React.SyntheticEvent<HTMLDivElement>) => void,
+    buttonState?: ButtonProps.ButtonState,
+    label?:string | object,
```
FrancescoCioria: you should replace `object` with `{ [key in buttonState]: string }`

src/button/Button.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    icon?:string | object,
```
FrancescoCioria: you should replace `object` with `{ [key in buttonState]: string }`

src/button/Button.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    children?: string,
+    type?: ButtonProps.ButtonType,
+    primary?: boolean,
+    flat?: boolean,
+    size?: ButtonProps.ButtonSize,
+    fluid?: boolean,
+    circular?: boolean,
+    textOverflow?: React.ComponentClass<object>, // todo(typo) when translate text overflow, update here its props types
+    style?: object,
```
FrancescoCioria: same as other PR: we should use `React.CSSProperties` for `style`

ascariandrea:
Fixed!

PR #951, opened 2017-06-28
#912: [typescript .d.ts] Panel (closes #912)

### LINKED ISSUE #912: [typescript .d.ts] Panel
← #910

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #912

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/Panel/Panel.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    size?: HeaderSize,
+    content?: any,
+    title?: ReactNode | string, // TODO(typo): wtf
+    hideTitleWhenExpanded?: boolean,
+    menu?: any
+  }
+}
+
+export type PanelProps = {
+  children: ReactNode,
```
FrancescoCioria: usually we use `any` for `children` but honestly I don't know exactly the reason...
@giogonzo is `ReactNode` ok?

src/Panel/Panel.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  type: 'docked-top' | 'docked-left' | 'docked-bottom' | 'docked-bottom' | 'floating',
+  header?: PanelProps.Header,
+  loading?: boolean,
+  dark?: boolean,
+  softLoading?: boolean,
+  softLoadingDelay?: boolean,
```
FrancescoCioria: this should be a `number`
(to be precise, a `NonNegativeNumber` but we can't check that statically with Typescript)

src/Panel/Panel.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  className?: string,
+  clearMargin?: 'top' | 'left' | 'bottom' | 'right',
+  style?: Object
```
FrancescoCioria: same: use `CSSProperties`

src/Panel/Panel.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  type Header = {
+    collapse?: {
+      direction: CollapseDirection,
+      onExpand: () => void,
+      onCollapse: () => void,
+      isCollapsed?: boolean
+    },
+    size?: HeaderSize,
+    content?: any,
+    title?: ReactNode | string, // TODO(typo): wtf
```
FrancescoCioria: mmm `ReactNode` should accept `string`s, can you check this again?

Also, why isn't this `any` like `content` and `menu`?

src/Panel/PanelHeader.d.ts
```diff
@@ -0,0 +1,18 @@
...
+// TODO: generalize with Panel.Props
+export type PanelHeaderProps = {
+  collapse?: {
+    direction: PanelProps.CollapseDirection,
+    onToggleExpanded: () => void,
+    isExpanded?: boolean
+  },
+  size?: PanelProps.HeaderSize,
+  content?: any,
+  title?: ReactNode | string, // TODO(typo): wtf
```
FrancescoCioria: same comment as `Panel`

src/Panel/PanelHeader.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  menu?: ReactChildren
```
FrancescoCioria: and this one `ReactChildren`?
I think there is a lot of confusion on how to translate `t.ReactChildren` in Typescript.

@giogonzo do we have a guideline on how to handle these cases?

src/Panel/Panel.d.ts
ascariandrea: nope, `ReactNode` doesn't accept a `string`
```
type ReactNode = ReactChild | ReactFragment | boolean | null | undefined;
```

but I'm going to set to `any` until we'll find a right type for `children`

ascariandrea:
Fixed!

PR #953, opened 2017-06-29
#940: [typescript .d.ts] Toggle (closes #940)

### LINKED ISSUE #940: [typescript .d.ts] Toggle
← #939

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #940

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/toggle/Toggle.d.ts
+++ b/src/toggle/Toggle.d.ts
@@ -1,0 +1,15 @@
+import { CSSProperties, PureComponent } from 'react'
+
+export type ToggleProps = {
+  value?: boolean,
+  onChange?: (value: boolean) => void,
+  valueLink?: {
+    value?: boolean,
+    requestChange: (value: boolean) => void
+  },
+  size?: number |string,
+  className?: string,
+  style?: CSSProperties
+}
+
+export default class Toggle extends PureComponent<ToggleProps, void> {}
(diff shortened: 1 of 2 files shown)


src/toggle/Toggle.d.ts
```diff
@@ -0,0 +1,15 @@
+import { CSSProperties, PureComponent } from 'react'
+
+export type ToggleProps = {
+  value?: boolean,
+  onChange: () => any, // TODO: not used
```
FrancescoCioria: this should be optional and with same signature as `requestChange`:
```
onChange?: (value: boolean) => void`
```

src/toggle/Toggle.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  valueLink?: {
+    value?: boolean,
+    requestChange: (value: boolean) => any
```
FrancescoCioria: it should return `void`

ascariandrea:
Done!

src/toggle/Toggle.d.ts
```diff
@@ -0,0 +1,15 @@
+import { CSSProperties, PureComponent } from 'react'
+
+export type ToggleProps = {
+  value?: boolean,
+  onChange?: () => void, // TODO: not used
```
FrancescoCioria: actually it's used, `getValueLink` returns and object with a `requestChange` that can be the original `props.onChange` or `props.valueLink.requestChange`

ascariandrea: 👍

PR #955, opened 2017-06-29
#915: [typescript .d.ts] Async Status Indicator, BrowserDetector, CollapsableSection, ConfirmationInput and Dropdown (closes #915)

### LINKED ISSUE #915: [typescript .d.ts] Async Status Indicator
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #915

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/browser-detector/BrowserDetector.d.ts
+++ b/src/browser-detector/BrowserDetector.d.ts
@@ -1,10 +1,11 @@
 import { PureComponent } from 'react';
+import { IBowserGrade } from '@types/bowser';
 
 type Browser = 'chrome' | 'firefox' | 'safari' | 'msie' | 'msedge' | 'opera';
 
 export type BrowserDetectorProps = {
   children: any, // TODO: t.ReactChildren
-  placeholder: (detectedBrowser: string) => any, // TODO: t.ReactChildren
+  placeholder: (detectedBrowser: IBowserGrade) => any, // TODO: t.ReactChildren
   supportedBrowsers?: Browser[],
   userAgent?: string
 }
--- a/src/browser-detector/BrowserDetector.d.ts
+++ b/src/browser-detector/BrowserDetector.d.ts
@@ -1,0 +1,12 @@
+import { PureComponent } from 'react';
+
+type Browser = 'chrome' | 'firefox' | 'safari' | 'msie' | 'msedge' | 'opera';
+
+export type BrowserDetectorProps = {
+  children: any, // TODO: t.ReactChildren
+  placeholder: (detectedBrowser: string) => any, // TODO: t.ReactChildren
+  supportedBrowsers?: Browser[],
+  userAgent?: string
+}
+
+export default class BrowserDetector extends PureComponent<BrowserDetectorProps, void> {}
--- a/src/browser-detector/BrowserDetector.d.ts
+++ b/src/browser-detector/BrowserDetector.d.ts
@@ -1,13 +1,19 @@
 import { PureComponent } from 'react';
-import { IBowserGrade } from '@types/bowser';
 
 type Browser = 'chrome' | 'firefox' | 'safari' | 'msie' | 'msedge' | 'opera';
 
+export interface DetectedBrowser {    // TODO: use bowser typings
+    name: string;
+    version: string|number;
+    osversion: string|number;
+}
+
+
 export type BrowserDetectorProps = {
   children: any, // TODO: t.ReactChildren
-  placeholder: (detectedBrowser: IBowserGrade) => any, // TODO: t.ReactChildren
+  placeholder: (detectedBrowser: DetectedBrowser) => any, // TODO: t.ReactChildren
   supportedBrowsers?: Browser[],
   userAgent?: string
 }
 
-export default class BrowserDetector extends PureComponent<BrowserDetectorProps, void> {}
+export default class BrowserDetector extends PureComponent<BrowserDetectorProps, {}> {}
--- a/src/dropdown/Dropdown.d.ts
+++ b/src/dropdown/Dropdown.d.ts
@@ -56,7 +56,7 @@ export type DropdownProps = {
   style?: CSSProperties
 }
 
-export default class Dropdown extends PureComponent<DropdownProps, void> {}
+export default class Dropdown extends PureComponent<DropdownProps> {}
 
 export var Props: {
   [key: string]: t.Type<any>
--- a/src/dropdown/Dropdown.d.ts
+++ b/src/dropdown/Dropdown.d.ts
@@ -6,7 +6,7 @@ type Value = number | string | Option | Option[];
 
 interface DropdownMenuRendererProps extends MenuRendererProps {
   instancePrefix: string,
-  onFocus: (e: SyntheticEvent<HTMLDivElement>) => void,
+  onFocus: FocusEventHandler<HTMLDivElement>,
   onSelect: (option: Option, e: SyntheticEvent<HTMLDivElement>) => void,
   groupByKey?: string,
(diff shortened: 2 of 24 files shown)


src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  valueKey: any
+}
+
+export type DropdownProps = {
+  value?: Value,
+  valueLink?: {
+    value?: Value,
+    requestChange: (value: Value) => void
+  },
+  onChange?: (value?: Value) => void,
```
FrancescoCioria: I'm not sure `value` can be a string o a number

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
+// TODO: use react-select types https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-select/index.d.ts
+
+import { CSSProperties, PureComponent, SyntheticEvent } from 'react'
+import * as t from 'tcomb';
+
+type Value = number | string | object | object[];
```
FrancescoCioria: Actually even better, we could use `@types/react-select` types here

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  disabled?: boolean,
+  searchable?: boolean,
+  clearable?: boolean,
+  backspaceRemoves?: boolean,
+  multi?: boolean,
+  flat?: boolean,
+  autoBlur?: boolean,
+  simpleValue?: boolean,
+  menuPosition?: 'top' | 'bottom',
+  menuRenderer?: (menuItem: object) => any, // TODO: t.ReactChildren
```
FrancescoCioria: here we could use `MenuRendererProps` type from `@types/react-select` + the two keys we add

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  groupByKey?: string,
+  optionGroupRenderer?: (title: string) => any, // TODO: t.ReactChildren
+  placeholder?: JSX.Element | string, // TODO: t.maybe(t.union([t.String, t.ReactElement]))
```
FrancescoCioria: better use any?

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  noResultsText?: JSX.Element | string,
```
FrancescoCioria: this should be only string

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  allowCreate?: boolean,
+  addLabelText?: JSX.Element | string,
```
FrancescoCioria: this should be only string

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  valueRenderer?: (option: object) => JSX.Element, // TODO: t.ReactChildren
```
FrancescoCioria: better return any?
also, I would improve the `object` type by using types from react-select

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  optionRenderer?: (option: object) => JSX.Element, // TODO: t.ReactChildren
```
FrancescoCioria: better return any?
also, I would improve the `object` type by using types from react-select

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  delimiter?: string,
+  onInputChange?: () => void,
```
FrancescoCioria: I'm pretty sure this function accepts at least one argument

src/popover/Popover.d.ts
```diff
@@ -38,3 +39,7 @@ export type PopoverProps = {
 }
 
 export default class Popover extends PureComponent<PopoverProps, void> {}
+
+export var Props: {
```
FrancescoCioria: why this?

src/dropdown/Dropdown.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  onCloseResetsInput?: boolean,
+  isLoading?: boolean,
+  id?: string,
+  className?: string,
+  style?: CSSProperties
+}
+
+export default class Dropdown extends PureComponent<DropdownProps, void> {}
+
+export var Props: {
```
FrancescoCioria: why this?

src/browser-detector/BrowserDetector.d.ts
```diff
@@ -0,0 +1,12 @@
+import { PureComponent } from 'react';
+
+type Browser = 'chrome' | 'firefox' | 'safari' | 'msie' | 'msedge' | 'opera';
+
+export type BrowserDetectorProps = {
+  children: any, // TODO: t.ReactChildren
+  placeholder: (detectedBrowser: string) => any, // TODO: t.ReactChildren
```
veej: https://github.com/lancedikson/bowser/blob/master/typings.d.ts doesn't seem to be very interesting :(

veej: woops... these are the real types, and they're more interesting :P https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bowser/index.d.ts

src/dropdown/Dropdown.d.ts
veej: `@types/react-select` defines it as `value?: Option | Option[] | string | string[] | number | number[] | boolean;`

src/dropdown/Dropdown.d.ts
veej: why `any`?

veej: oh, `@types/react-select` uses `placeholder?: string | React.ReactElement<any>`... I think we should use the same

src/browser-detector/BrowserDetector.d.ts
veej: Since `bowser` is defining its own types, it seems there's no way to use the ones from `@types/bowser` [](https://github.com/lancedikson/bowser/issues/182)
I cut&paste the relevant interface

src/dropdown/Dropdown.d.ts
veej: Needed since `dropdown.js` is exporting tcomb `Props`

src/dropdown/Dropdown.d.ts
```diff
@@ -39,19 +35,19 @@ export type DropdownProps = {
...
-  optionRenderer?: (option: object) => JSX.Element, // TODO: t.ReactChildren
+  addLabelText?: string,
+  valueRenderer?: (option: Option) => JSX.Element, // TODO: t.ReactChildren
+  optionRenderer?: (option: Option) => JSX.Element, // TODO: t.ReactChildren
   delimiter?: string,
-  onInputChange?: () => void,
-  onFocus?: () => void,
-  onBlur?: () => void,
+  onInputChange?: (inputValue: string) => void,
+  onFocus?: FocusEventHandler<{}>,
```
FrancescoCioria: I know you copy-pasted this from `@types/react-select`, but I think we should pass an HTML element to `FocusEventHandler` as we do for every other `SyntheticEvent` like: `FocusEventHandler<HTMLDivElement>` (if it's a div or simply `HTMLElement` in the other case)

PR #956, opened 2017-06-30
#219: implement `DateInputs` (better name?) (closes #219)

### LINKED ISSUE #219: implement `DateInputs` (better name?)
on Hailadoc (https://github.com/buildo/hailadoc/issues/1646) @ascariandrea made a first working example of this component but, as we were on a hurry, we didn't carefully define requirements nor specs. The result is a working component but not general enough (nor safe, I guess?) to be moved on this repo.
So, let's define requirements and specs :)
## requirements
- should have the standard API for form fields: `value`, `onChange`, `valueLink`
- should work out of the box with `tcomb-form`
- if possible, we'd better avoid using `momentJS`
  - find a replacement for date validation...
- user passes and receives only one value (specs might be: `Date`, `null` or `Error`), he shouldn't care about `day`, `month`, `year`
- labels (**TODO**)
  - general label?
  - label for each input?
- status (**TODO**):
  - valid
  - on focus
  - not valid (single input validation too?)

FrancescoCioria:
Closes #219

### tests performed
tested with the two examples:
- user inputs is never overwritten by the new `props.value`
- validation seems to works

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/date-field/DateField.d.ts
+++ b/src/date-field/DateField.d.ts
@@ -1,4 +1,5 @@
 import { CSSProperties, ComponentType } from 'react';
+import { Type } from 'tcomb';
 
 export type DateFieldProps = {
   value?: Date,
@@ -11,6 +12,10 @@ export type DateFieldProps = {
   },
   inputTypeNumber?: boolean
 };
+
+export const Props: {
+  [key: string]: Type<any>
+}
 
 declare const DateField: ComponentType<DateFieldProps>;
 export default DateField;
--- a/src/date-field/DateField.d.ts
+++ b/src/date-field/DateField.d.ts
@@ -1,0 +1,16 @@
+import { CSSProperties, ComponentType } from 'react';
+
+export type DateFieldProps = {
+  value?: Date,
+  onChange: (value: Date) => void,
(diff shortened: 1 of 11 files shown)


src/date-field/DateField.js
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
+import { props, t } from '../utils';
+import every from 'lodash/every';
+import cx from 'classnames';
+import View from 'react-flexview';
+
+import './dateField.scss';
```
FrancescoCioria: ups! great 👁

src/date-field/DateField.d.ts
```diff
@@ -0,0 +1,16 @@
...
+  placeholders?: {
+    day?: string,
+    month?: string,
+    year?: string
+  },
+  inputTypeNumber?: boolean
+};
+
+declare const DateField: ComponentType<DateFieldProps>;
+export default DateField;
```
veej: The component is exporting `Props`. We should add the usual definition here:

```
export const Props: {
  [key: string]: Type<any>
}
```

FrancescoCioria: done

PR #958, opened 2017-06-30
#914: [typescript .d.ts] TextOverflow, Button, StatefulButton, Panel, PanelHeader, TabbedPanel and Tooltip (closes #914)

### LINKED ISSUE #914: [typescript .d.ts] Tooltip
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #914

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/Panel/Panel.d.ts
+++ b/src/Panel/Panel.d.ts
@@ -1,20 +1,17 @@
 import { CSSProperties, PureComponent } from 'react';
 import { Type } from 'tcomb';
+import { PanelHeaderProps } from './PanelHeader';
 
 export namespace PanelProps {
 
-  type HeaderSize = 'tiny' | 'small' | 'medium';
-
-  type CollapseDirection = 'up' | 'left' | 'down' | 'right';
-
   type Header = {
     collapse?: {
-      direction: CollapseDirection,
+      direction: 'up' | 'left' | 'down' | 'right',
       onExpand: () => void,
       onCollapse: () => void,
       isCollapsed?: boolean
     },
-    size?: HeaderSize,
+    size?: PanelHeaderProps.HeaderSize,
     content?: any, // TODO: t.ReactChildren
     title?: any, // TODO(typo): wtf
     hideTitleWhenExpanded?: boolean,
@@ -29,7 +26,7 @@ export type PanelProps = {
   loading?: boolean,
   dark?: boolean,
   softLoading?: boolean,
-  softLoadingDelay?: number,
+  softLoadingDelay?: number, // TODO: should be non negative
   className?: string,
   clearMargin?: 'top' | 'left' | 'bottom' | 'right',
   style?: CSSProperties
--- a/src/Panel/Panel.d.ts
+++ b/src/Panel/Panel.d.ts
@@ -1,4 +1,5 @@
 import { CSSProperties, PureComponent } from 'react';
+import { Type } from 'tcomb';
 
 export namespace PanelProps {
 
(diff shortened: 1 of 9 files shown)


src/button/Button.d.ts
```diff
@@ -21,9 +22,17 @@ export type ButtonProps = {
     fluid?: boolean,
     circular?: boolean,
     textOverflow?: React.ComponentClass<object>, // todo(typo) when translate text overflow, update here its props types
```
francescogior: also here ComponentType, and you may pass Textoverflow's props now

src/button/Button.d.ts
```diff
@@ -21,9 +22,17 @@ export type ButtonProps = {
...
-    style?: React.CSSProperties,
+    style?: CSSProperties,
     className?: string
   }
```
francescogior: what about StatefulButtonProps?
they should be `ButtonProps & { baseState?: BaseState, stableSuccess?: boolean, timerMillis?: number }`

francescogior: also, there where tons of validations (`propsInvariants`), probably impossible to replicate in TS, but I'd add a TODO as usual

src/button/Button.d.ts
```diff
@@ -21,9 +22,17 @@ export type ButtonProps = {
...
-export default class Button extends React.PureComponent<ButtonProps> {}
+export default class Button extends PureComponent<ButtonProps> {}
+
+export const Props: {
+  [key: string]: Type<any>
+}
+
+export const buttonStates: ButtonProps.ButtonState;
+export const buttonSizes: ButtonProps.ButtonSize;
+export const buttonTypes: ButtonProps.ButtonType;
```
francescogior: the three above are arrays

```js
export const buttonStates: ButtonProps.ButtonState[];
export const buttonSizes: ButtonProps.ButtonSize[];
export const buttonTypes: ButtonProps.ButtonType[];
```

src/Panel/Panel.d.ts
```diff
@@ -35,3 +36,7 @@ export type PanelProps = {
 }
```
francescogior: why `docked-top' | ...` is not declared in the namespace like the others?

src/Tooltip/Tooltip.d.ts
```diff
@@ -0,0 +1,24 @@
...
+import { PopoverProps } from '../popover/Popover';
+
+export namespace TooltipProps {
+  type Type = 'light' | 'dark';
+  type Size = 'small' | 'big';
+}
+
+export type TooltipProps = {
+  children: any, // TODO: t.ReactChildren
+  popover: PopoverProps,
```
francescogior: add TODO, was refinement 
`p => t.String.is(p.content) && t.Nil.is(p.event))`

src/text-overflow/TextOverflow.d.ts
```diff
@@ -1,13 +1,18 @@
 import { CSSProperties, PureComponent } from 'react';
+import { Type } from 'tcomb';
 
 export type TextOverflowProps = {
   children?: (self: JSX.Element, isOpen: boolean) => any,
   label: string,
+  popover?: object, // TODO(typo): Omit?
```
FrancescoCioria: @ascariandrea as I told you in the other PRs, this should reflect exactly the `@props` defined in this repo.

Please don't copy-paste from alinity-pro, those are the props of the custom wrappers.
If you start from those you will always make mistakes as they are a different thing :)

ascariandrea: I think I made a wrong rebase :/

src/button/Button.d.ts
ascariandrea: > what about StatefulButtonProps?

They're defined in `StatefulButton.d.ts`

src/Panel/Panel.d.ts
ascariandrea: I've changed the exports to replicate the originals

PR #959, opened 2017-07-04
#910: [typescript .d.ts] DropdownMenu (closes #910)

### LINKED ISSUE #910: [typescript .d.ts] DropdownMenu
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #910

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/DropdownMenu/DropdownMenu.d.ts
+++ b/src/DropdownMenu/DropdownMenu.d.ts
@@ -1,9 +1,9 @@
-import { PureComponet } from 'react';
+import { PureComponent } from 'react';
 import { MenuProps } from './Menu';
 
 export type DropdownMenuProps = {
   children?: any, // TODO: t.ReactChildren
-  menuRenderer?: (options: MenuProps.Option) => JSX.Element,
+  menuRenderer?: (options: MenuProps.Option[]) => JSX.Element,
   options: MenuProps.Option[],
   iconClassName?: string,
   isOpen?: boolean,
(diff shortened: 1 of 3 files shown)


FrancescoCioria:
@ascariandrea can you rebase? (there are 17 changed files in this PR but I suppose most of them are already like this on `master`)

ascariandrea:
@FrancescoCioria done!

src/DropdownMenu/DropdownMenu.d.ts
```diff
@@ -0,0 +1,18 @@
+import { PureComponet } from 'react';
+import { MenuProps } from './Menu';
+
+export type DropdownMenuProps = {
+  children?: any, // TODO: t.ReactChildren
+  menuRenderer?: (options: MenuProps.Option) => JSX.Element,
```
FrancescoCioria: this is a list of options: `MenuProps.Option[]`

src/DropdownMenu/index.d.ts
```diff
@@ -0,0 +1,3 @@
+import DropdownMenu from './DropdownMenu';
```
FrancescoCioria: minor: If I'm not mistaken, when using only `export` you can do `export X from './X` (it doesn't work only for `export default`)

src/DropdownMenu/DropdownMenu.d.ts
```diff
@@ -0,0 +1,18 @@
+import { PureComponet } from 'react';
```
FrancescoCioria: typo! 🚨

PR #960, opened 2017-07-04
#913: [typescript .d.ts] ScrollView (closes #913)

### LINKED ISSUE #913: [typescript .d.ts] ScrollView
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

francescogior:
Closes #913

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/ScrollView/ScrollView.d.ts
+++ b/src/ScrollView/ScrollView.d.ts
@@ -1,7 +1,7 @@
 import { CSSProperties, PureComponent, ComponentType } from 'react';
 import { Type } from 'tcomb';
 
-export type ScrollViewProps<CP> = {
+export type ScrollViewProps<CP> = { // TODO(typo) not really working
   children: any,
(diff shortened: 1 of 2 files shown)


src/ScrollView/ScrollView.d.ts
```diff
@@ -0,0 +1,18 @@
+import { CSSProperties, PureComponent, ComponentType } from 'react';
+import { Type } from 'tcomb';
+
+export type ScrollViewProps = {
+  children: any,
+  autoshow?: boolean,
+  forceGemini?: boolean,
+  component?: string | ComponentType,
```
ascariandrea: Shouldn't `Component` need a type parameter?

francescogior: Not sure if it defaults to `any` or if it's necessary

PR #962, opened 2017-07-04
#936: [typescript .d.ts] Tablo (closes #936)

### LINKED ISSUE #936: [typescript .d.ts] Tablo
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

francescogior:
Closes #936


src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  backgroundColor?: React.CSSProperties['backgroundColor'],
+  color?: React.CSSProperties['color'],
+  vAlignContent?: 'top' | 'center' | 'bottom',
+  hAlignContent?: 'left' | 'center' | 'right',
+  contentStyle?: React.CSSProperties,
+  style?: React.CSSProperties
+};
+
+export type ColumnProps = {
+  name: string,
```
giogonzo: this could be `ColumnProps<T>` and this `name: keyof T`

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  children?: any, // TODO(typo) it should be (Header | Cell | Footer)[],
+  allowCellsRecycling?: boolean
+};
+
+type SelectedRows = string[];
+type SortBy = string;
+type SortDir = 'asc' | 'desc';
+
+export type TabloProps = {
+  data: any[],
```
giogonzo: same here, `T[]`

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  className?: string,
+  rowHeight?: number,
+  headerHeight?: number,
+  groupHeaderHeight?: number,
+  footerHeight?: number,
+  onRowMouseEnter?: (rowIndex: number) => void,
+  onRowMouseLeave?: (rowIndex: number) => void,
+  columnsOrder?: string[],
```
giogonzo: and here, `(keyof T)[]`

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  height?: undefined,
+  autosize?: true
+});
+
+export const Header: React.StatelessComponent<HeaderProps>;
+export const Footer: React.StatelessComponent<FooterProps>;
+export const Cell: React.StatelessComponent<CellProps>;
+export const ColumnGroup: React.StatelessComponent<ColumnGroupProps>;
+export const Column: React.StatelessComponent<ColumnProps>;
+export default class Tablo extends React.PureComponent<TabloProps> {}
```
giogonzo: and this could be
```tsx
declare const Tablo: React.ComponentClass<TabloProps>;
export default Tablo;
```
not sure which one we prefer

francescogior:
@giogonzo the `T` type parameter you are suggesting would be an object/interface.

While it's fine to type `Tablo`'s`data` (`T` could be `API_ICQ_QCAnalysis`), `Column`'s `name` (and then `columnsOrder`) are not `keyof T`. They could be in some cases, but it's not in all our grids.
For example, we have column `workcell` while `keyof API_ICQ_QCAnalysis` contains `_workcellSerialNumber` and `workcellName`.

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  height?: undefined,
+  autosize?: true
+});
+
+export const Header: React.StatelessComponent<HeaderProps>;
+export const Footer: React.StatelessComponent<FooterProps>;
+export const Cell: React.StatelessComponent<CellProps>;
+export const ColumnGroup: React.StatelessComponent<ColumnGroupProps>;
+export const Column: React.StatelessComponent<ColumnProps>;
+declare const Tablo: React.ComponentClass<TabloProps>;
```
ascariandrea: Why a `const` instead of a `class`?

francescogior: It's just an alternative way to export the component @FrancescoCioria and @giogonzo told me about it.
It avoids to use `(Pure)Component` and the mismatch with the old and new react typing (state default to `{}`)

src/tablo/Tablo.d.ts
```diff
@@ -76,8 +76,8 @@ export type TabloProps<Data, Columns> = {
 
 export const Header: React.StatelessComponent<HeaderProps>;
 export const Footer: React.StatelessComponent<FooterProps>;
-export const Cell: React.StatelessComponent<CellProps>;
+export const Cell: React.StatelessComponent<CellProps<any>>;
 export const ColumnGroup: React.StatelessComponent<ColumnGroupProps>;
-export const Column: React.StatelessComponent<ColumnProps>;
-declare const Tablo: React.ComponentClass<TabloProps>;
+export const Column: React.StatelessComponent<ColumnProps<string>>
+declare const Tablo: React.ComponentClass<TabloProps<any, string>>;
```
FrancescoCioria: why is `Columns` a `string`?
shouldn't it be `string[]` or be called `Column`?
or is it really a string with a separator like `"columnA, columnB"`?

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react';
+
+export type HeaderProps = {
```
ascariandrea: I would namespace this

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  children: any
+};
+
+export type FooterProps = {
```
ascariandrea: This one too

src/tablo/Tablo.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  className?: string,
+  rowHeight?: number,
+  headerHeight?: number,
+  groupHeaderHeight?: number,
+  footerHeight?: number,
+  onRowMouseEnter?: (rowIndex: number) => void,
+  onRowMouseLeave?: (rowIndex: number) => void,
+  columnsOrder?: Columns[],
+  onColumnsReorder?: (columns: string[]) => void,
+  onColumnResize?: (x: { width: number, key: string}) => void,
```
ascariandrea: `onColumnResize?: (x: { width: number, key: string}) => void,`
could be 
`onColumnResize?: (x: { width: number, key: keyof Columns }) => void,`
(I think)

src/tablo/Tablo.d.ts
```diff
@@ -47,8 +47,8 @@ export type TabloProps<Data, Columns> = {
   onRowMouseEnter?: (rowIndex: number) => void,
   onRowMouseLeave?: (rowIndex: number) => void,
   columnsOrder?: Columns[],
-  onColumnsReorder?: (columns: string[]) => void,
```
FrancescoCioria: again, why plural `Columns` instead of `Column`?

src/tablo/Tablo.d.ts
francescogior: because it should be an enum: `'column1' | 'column2' | 'column3' ...`
I pushed a commit where I replaced some usages of `string` with `Columns` (just in order to document better)

src/tablo/Tablo.d.ts
francescogior: `onColumnsReorder` take the entire list of ordered columns as argument

francescogior: ok, it may be called `Column` instead of `Columns`

src/tablo/Tablo.d.ts
francescogior: already changed, using directly `Columns` since it's ideally an enum (the list of the columns of the table), but I typed it as `string` in the export.

src/tablo/Tablo.d.ts
```diff
@@ -1,41 +1,45 @@
 import * as React from 'react';
 
-export type HeaderProps = {
-  children: any
-};
+export namespace Tablo {
```
ascariandrea: You should export as `TabloProps`

src/tablo/Tablo.d.ts
FrancescoCioria: otherwise when I read `Columns[]` I think of a list of a list of columns :)

PR #967, opened 2017-07-07
#923: [typescript .d.ts] FocusableView (closes #923)

### LINKED ISSUE #923: [typescript .d.ts] FocusableView
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Closes #923

### tests performed
tested on AlinityPRO:
- `FocusableView` is typed
- `focusable-view/FocusableView` exports both the component and `FocusableViewProps`

### DIFF
--- a/src/focusable-view/FocusableView.d.ts
+++ b/src/focusable-view/FocusableView.d.ts
@@ -1,0 +1,22 @@
+import { CSSProperties, ComponentClass, ComponentType, FocusEventHandler } from 'react';
+import { Type } from 'tcomb';
+
+export type FocusableViewProps = {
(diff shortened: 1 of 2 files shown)


src/focusable-view/FocusableView.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  className?: string,
+  style?: CSSProperties,
+  [k: string]: any
+};
+
+export const Props: {
+  [key: string]: Type<any>
+};
+
+declare const FocusableView: ComponentClass<FocusableViewProps>;
```
ascariandrea: I've exported all previous components as `export default class Component extends PureComponent<ComponentProps> {}`, so we should define a standard method for this.

PR #969, opened 2017-07-07
#911: [typescript .d.ts] Icon (closes #911)

### LINKED ISSUE #911: [typescript .d.ts] Icon
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Closes #911

### tests performed
```js
import Icon from 'buildo-react-components/lib/icon';
import I, { IconProps, Props } from 'buildo-react-components/lib/icon/Icon';
```
No errors!

Also, tried some props -> correctly typed

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/Icon/Icon.d.ts
+++ b/src/Icon/Icon.d.ts
@@ -1,0 +1,18 @@
+import { CSSProperties, ComponentClass, MouseEventHandler } from 'react';
+import { Type } from 'tcomb';
+
+export type IconProps = {
+  icon?: string,
+  color?: string,
+  paths?: number, // it was PositiveInteger
+  onClick?: MouseEventHandler<HTMLElement>,
+  className?: string,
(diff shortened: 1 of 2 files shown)


src/Icon/Icon.d.ts
```diff
@@ -0,0 +1,18 @@
+import { CSSProperties, ComponentClass, MouseEventHandler } from 'react';
+import { Type } from 'tcomb';
+
+export type IconProps = {
+  icon?: string,
+  color?: string,
+  paths?: number, // it was PositiveInteger
+  onClick?: MouseEventHandler<HTMLElement>,
```
FrancescoCioria: I checked: there is no `HTMLIElement`. internally react uses generic `HTMLElement` for `<i>`

PR #973, opened 2017-07-07
#972: POC for new DatePicker (closes #972)

### LINKED ISSUE #972: POC for new DatePicker
## requirements
Since `rc-datepicker` is missing important features and maintaining it is becoming more and more time consuming, we'd like to implement a new `DatePicker` wrapping a more famous one, like `react-dates`

## specs
- This POC is an attempt to maintain the same old API wrapping a component like `react-dates`.
- Month/Year fast navigation is not strictly required. Picking dates in the past/future (like birth dates) will be covered by a different, more appropriate component.

## misc
{optional: other useful info}

veej:
Closes #972


package.json
```diff
@@ -48,6 +48,8 @@
     "linkify-it": "^2.0.3",
     "lodash": "^4.6.1",
     "mobile-detect": "^1.2.1",
+    "react-addons-shallow-compare": "^15.6.0",
```
FrancescoCioria: is this really used?

src/DatePicker/DatePicker.js
```diff
@@ -0,0 +1,215 @@
+import React from 'react';
+import cx from 'classnames';
+import { SingleDatePicker } from 'react-dates';
+import FlexView from 'react-flexview';
+import { props, t, skinnable } from '../utils';
+import moment from 'moment';
```
FrancescoCioria: if we need `moment` we should add it to the deps

src/DatePicker/README.md
```diff
@@ -0,0 +1,14 @@
+# NewDatePicker
```
FrancescoCioria: I think this is outdated, try running `npm run generate-all-readmes`

src/DatePicker/examples/default.js
```diff
@@ -0,0 +1,26 @@
+class Example extends React.Component {
+
+  constructor() {
```
FrancescoCioria: minor: replace `constructor` with cleaner `state = {}`

FrancescoCioria:
@veej also could you add the `d.ts` typings?

package.json
veej: Unfortunately yes, it's needed by [react-dates](https://github.com/airbnb/react-dates) (see `Install dependencies` section)

PR #974, opened 2017-07-07
#938: [typescript .d.ts] Time Picker (closes #938)

### LINKED ISSUE #938: [typescript .d.ts] Time Picker
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Closes #938

### tests performed
```js
import TimePicker from 'buildo-react-components/lib/time-picker';
import T, { TimePickerProps, Props } from 'buildo-react-components/lib/time-picker/TimePicker';
```

No errors!

Also, tried some props -> correctly typed

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/time-picker/TimePicker.d.ts
+++ b/src/time-picker/TimePicker.d.ts
@@ -1,5 +1,5 @@
 import { CSSProperties, ComponentClass, MouseEventHandler } from 'react';
-import { Type, Struct } from 'tcomb';
+import { Type } from 'tcomb';
 import { Option, MenuRendererProps } from 'react-select';
 
 type TimeFormat = '24h' | '12h';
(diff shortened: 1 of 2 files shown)


src/time-picker/TimePicker.d.ts
```diff
@@ -0,0 +1,45 @@
+import { CSSProperties, ComponentClass, MouseEventHandler } from 'react';
+import { Type, Struct } from 'tcomb';
```
ascariandrea: `Struct` seems unused

FrancescoCioria: 👍

PR #981, opened 2017-07-10
#980: Fix tablo (closes #980)

### LINKED ISSUE #980: Fix tablo

francescogior:
Closes #980

### DIFF
--- a/src/tablo/plugins/columnsReorder/columnsReorderGrid.js
+++ b/src/tablo/plugins/columnsReorder/columnsReorderGrid.js
@@ -25,7 +25,6 @@ export default (Grid) =>
     columnsOrder: maybe(list(t.String)),
     onColumnsReorder: maybe(t.Function)
   }, { strict: false })
-  // @skinnable(contains(Grid))
   class ColumnsReorderGrid extends React.PureComponent {
 
     constructor() {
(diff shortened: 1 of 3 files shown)


giogonzo:
ok we have something now.. the real task here is to discover how react-skinnable is broken with respect to revenge/skinnable then?

src/tablo/plugins/columnsReorder/columnsReorderGrid.js
```diff
@@ -25,7 +25,7 @@ export default (Grid) =>
     columnsOrder: maybe(list(t.String)),
     onColumnsReorder: maybe(t.Function)
   }, { strict: false })
-  @skinnable(contains(Grid))
+  // @skinnable(contains(Grid))
```
FrancescoCioria: maybe remove completely instead of commenting? (as you've done in other files)

FrancescoCioria:
@francescogior can you briefly explain me why `react-skinnable` was breaking sorting and reorder?

francescogior:
> @francescogior can you briefly explain me why react-skinnable was breaking sorting and reorder?

I didn't find out yet. @ascariandrea is also investigating

giogonzo:
ups sorry, I missed the approval.
It means we decided to avoid using skinnable here, and to investigate further in the longer term. Is there an issue for this? Can you please open one if not, @francescogior ?

Also, whatever it is, it could break every other component using skinnable in obscure ways... Should we remove it everywhere in the meantime @FrancescoCioria ?

PR #982, opened 2017-07-10
#934: [typescript .d.ts] ResizeSensor (closes #934)

### LINKED ISSUE #934: [typescript .d.ts] ResizeSensor
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #934

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/resize-sensor/ResizeSensor.d.ts
+++ b/src/resize-sensor/ResizeSensor.d.ts
@@ -3,7 +3,7 @@ import { Type } from 'tcomb';
 
 export type ResizeSensorProps = {
   children: any,
-  onResize: (elements: HTMLElement) => void,
+  onResize: () => void,
   debounce?: number // TODO: integer
 }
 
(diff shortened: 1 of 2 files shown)


src/resize-sensor/ResizeSensor.d.ts
```diff
@@ -0,0 +1,14 @@
+import { PureComponent } from 'react';
+import { Type } from 'tcomb';
+
+export type ResizeSensorProps = {
+  children: any,
+  onResize: (elements: HTMLElement) => void,
```
FrancescoCioria: I think `onResize` is called with no argument

ascariandrea: I digged a little into `css-element-queries` and found https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js#L52

FrancescoCioria: yeah, but your should look at `this.props.onResize` inside the component and right now it's called with no arg

ascariandrea: 👍 fixed!

PR #983, opened 2017-07-10
#928: [typescript .d.ts] MobileDetector (closes #928)

### LINKED ISSUE #928: [typescript .d.ts] MobileDetector
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #928

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/mobile-detector/MobileDetector.d.ts
+++ b/src/mobile-detector/MobileDetector.d.ts
@@ -1,0 +1,21 @@
+import { PureComponent } from 'react';
+import { Type } from 'tcomb';
+
+type ChildrenArgs = {
+  isDesktop: boolean,
+  isMobile: boolean,
+  isPhone: boolean,
+  isTablet: boolean
+};
+
+export type MobileDetectorProps = {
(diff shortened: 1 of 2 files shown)


src/mobile-detector/MobileDetector.d.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type ChildrenArgs = {
+  isDesktop: boolean,
+  isMobile: boolean,
+  isPhone: boolean,
+  isTablet: boolean
+};
+
+export type MobileDetectorProps = {
+  children: ({ isDesktop, isMobile, isPhone, isTablet }: ChildrenArgs) => JSX.Element
```
francescogior: just as a style convention, I think the destructuring is useless here
```js
children: (childrenArgs: ChildrenArgs) => JSX.Element
```

ascariandrea: You're right, give me a minute

ascariandrea: Done!

PR #984, opened 2017-07-10
#929: [typescript .d.ts] ModalManager (closes #929)

### LINKED ISSUE #929: [typescript .d.ts] ModalManager
← #908

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #929

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/modal-manager/ModalManager.d.ts
+++ b/src/modal-manager/ModalManager.d.ts
@@ -1,0 +1,18 @@
+import { PureComponent } from 'react';
+import { Type } from 'tcomb';
+import { TransitionWrapperProps } from '../transition-wrapper/TransitionWrapper';
+
+export type ModalManagerProps = {
+  children?: any,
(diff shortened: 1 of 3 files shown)


src/modal-manager/ModalManager.d.ts
```diff
@@ -0,0 +1,17 @@
+import { CSSProperties, PureComponent } from 'react';
+import { Type } from 'tcomb';
+
+export type ModalManagerProps = {
+  children?: any,
+  transitionStyles?: { [key: string]: CSSProperties },
```
FrancescoCioria: this could be better typed as `TransitionWrapperProps .TransitionStyles` from `TransitionWrapper.d.ts`

ascariandrea: Fixed!

PR #988, opened 2017-07-12
Popover position auto

ascariandrea:
I added an example to test the correct behaviour

### DIFF
--- a/src/popover/Popover.js
+++ b/src/popover/Popover.js
@@ -11,7 +11,7 @@ export const Props = {
   popover: t.struct({
     content: t.ReactChildren,
     attachToBody: t.maybe(t.Boolean),
-    position: t.maybe(t.enums.of(['top', 'bottom', 'left', 'right'])),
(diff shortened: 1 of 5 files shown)


src/popover/Popover.js
```diff
@@ -11,7 +11,7 @@ export const Props = {
   popover: t.struct({
     content: t.ReactChildren,
     attachToBody: t.maybe(t.Boolean),
-    position: t.maybe(t.enums.of(['top', 'bottom', 'left', 'right'])),
+    position: t.maybe(t.enums.of(['top', 'bottom', 'left', 'right', 'auto'])),
```
FrancescoCioria: As you're trying to compute both `position` and `anchor` I would add a new prop `auto: t.maybe(t.Boolean)` that should exist only if no `position` and `anchor` are passed

FrancescoCioria:
@ascariandrea PPS: I pushed a commit that fixes style of the examples for absolute popovers (`attachToBody: true`)

ascariandrea:
I would close this in favor of #992

PR #992, opened 2017-07-12
Popover anchor auto

ascariandrea:
See examples :)

### DIFF
--- a/src/popover/Popover.js
+++ b/src/popover/Popover.js
@@ -412,32 +412,44 @@ export default class Popover extends React.Component {
         anchorOffset.top = isVertical ? (child.height - popover.height) : 0;
         break;
       case 'auto':
-        _realAnchor = 'start';
+        _realAnchor = 'center';
         if (isHorizontal) {
           const scrollX = (window.pageXOffset || document.scrollLeft || 0) - (document.clientLeft || 0);
-          let popoverXBoundary = popover.x + popover.width + distance;
+          // center the popover horizontally as first step
+          anchorOffset.left = (child.width - popover.width) / 2;
+
+          // check its x boundary doesn't overflow on right
+          const popoverXBoundary = child.x + anchorOffset.left + popover.width + distance;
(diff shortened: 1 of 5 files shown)


src/popover/Popover.js
```diff
@@ -410,6 +411,36 @@ export default class Popover extends React.Component {
         anchorOffset.left = isHorizontal ? (child.width - popover.width) : 0;
         anchorOffset.top = isVertical ? (child.height - popover.height) : 0;
         break;
+      case 'auto':
+        _realAnchor = 'start';
+        if (isHorizontal) {
+          const scrollX = (window.pageXOffset || document.scrollLeft || 0) - (document.clientLeft || 0);
```
FrancescoCioria: what is `scrollX` exactly? does it work also when the popover is inside a `div` that has `overflow: auto`?

src/popover/Popover.js
```diff
@@ -410,6 +411,36 @@ export default class Popover extends React.Component {
         anchorOffset.left = isHorizontal ? (child.width - popover.width) : 0;
         anchorOffset.top = isVertical ? (child.height - popover.height) : 0;
         break;
+      case 'auto':
+        _realAnchor = 'start';
```
FrancescoCioria: it would be great if the default remained `center` to avoid changing the UI of AlinityPRO.
do you think it would be more complicated?

ascariandrea: Yep, cause we should add logic to handle the case where it overflows the left side of the body.

src/popover/Popover.js
ascariandrea: The scroll of the `document` (otherwise we would have also the element's scrollLeft as parameter of the operation)

ascariandrea:
Fixed by setting `center` as default

FrancescoCioria:
seems to work perfectly!
I'll try it quickly on AlinityPRO to see if it fixes the original defect :)

ascariandrea:
It does 😎

PR #995, opened 2017-07-13
#993: Popover should handle context  (closes #993)

### LINKED ISSUE #993: Popover should handle context
## requirements
When `attachToBody` is passed to `Popover` it creates a new render tree without passing the received context. 

## specs
- add a prop to `Popover` to describe the context
- wrap `Popover`'s children in a component responsible for assigning the context

## misc
{optional: other useful info}

ascariandrea:
Closes #993

### tests performed
- tested on alinity-pro ✅ 

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/popover/Popover.js
+++ b/src/popover/Popover.js
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import cx from 'classnames';
 import debounce from 'lodash/debounce';
-import { props, t } from '../utils';
+import { props, t, getContextWrapper } from '../utils';
 
 const NO_SIZE_WRAPPER = 'no-size-wrapper';
 
@@ -30,12 +30,15 @@ export const Props = {
     delay: t.maybe(t.union([
       t.Integer,
       t.interface({ whenClosed: t.maybe(t.Integer), whenOpen: t.maybe(t.Integer) })
-    ]))
+    ])),
+    contextTypes: t.maybe(t.Object),
+    context: t.maybe(t.Object)
   }),
   id: t.maybe(t.String),
   className: t.maybe(t.String),
   style: t.maybe(t.Object)
 };
+
 
 /**
  * Composed of two children: trigger (children) and popover. After a particular event on the trigger (usually "hover" or "click") it renders the popover and positions it relative to it.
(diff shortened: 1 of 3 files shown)


src/popover/Popover.js
```diff
@@ -69,7 +72,8 @@ export default class Popover extends React.Component {
   componentDidUpdate() {
     if (this.containerNode) {
       const popover = this.getVisiblePopover();
-      ReactDOM.render(popover, this.containerNode);
+      const ContextWrapper = getContextWrapper(this.props.popover.contextTypes);
```
FrancescoCioria: you should use `this.getPopoverProps()` instead of `this.props`

src/popover/Popover.js
```diff
@@ -69,7 +72,8 @@ export default class Popover extends React.Component {
...
+      ReactDOM.render(<ContextWrapper context={this.props.popover.context}>{popover}</ContextWrapper>, this.containerNode);
```
FrancescoCioria: here `context` is optional but in `ContextWrapper` is required

src/popover/Popover.js
```diff
@@ -236,7 +240,8 @@ export default class Popover extends React.Component {
 
     // render invisible popover
     const hiddenPopover = this.getHiddenPopover();
-    ReactDOM.render(hiddenPopover, this.containerNode);
+    const ContextWrapper = getContextWrapper(this.props.popover.contextTypes);
+    ReactDOM.render(<ContextWrapper context={this.props.popover.context}>{hiddenPopover}</ContextWrapper>, this.containerNode);
```
FrancescoCioria: same

src/popover/examples/hover-auto-attach-to-body.js
```diff
@@ -0,0 +1,38 @@
+const Option = ({ title, value }) => (
```
FrancescoCioria: this file shouldn't be here 😛

src/utils/index.js
```diff
@@ -1,5 +1,27 @@
+import React from 'react';
 import cx from 'classnames';
+import { props, t } from 'tcomb-react';
 
 export skinnable, { contains } from 'react-skinnable';
-export { props, t } from 'tcomb-react';
+export { props, t };
 export const stateClassUtil = (...classes) => cx([].concat(...classes).map(cl => `is-${cl}`));
+
+export const getContextWrapper = (contextTypes) => {
```
giogonzo: minor, but the same exact thing I think is already implemented for Modal here: https://github.com/buildo/react-components/blob/master/src/modal/modalWithContext.js
What about reusing/sharing?

giogonzo: no, probably not exactly the same.. if not straightforward, 👍 as it is

src/popover/examples/hover-auto-attach-to-body.js
ascariandrea: Removed 🙈

src/utils/index.js
ascariandrea: 👍

PR #998, opened 2017-07-14
#492: Components folders should all follow the same name-convention (closes #492)

### LINKED ISSUE #492: Components folders should all follow the same name-convention
## requirements

components folder should follow a precise name-convention and devs should be able to guess the component folder name by knowing the component name.
## specs
- use `ComponentCase` naming (the one we're using in AlinIQ)
  - ex: `src/text-overflow/TextOverflow.js` --> **`src/TextOverflow/TextOverflow.js`**
## misc

{optional: other useful info}

veej:
Closes #492

### tests performed
- showroom is working for each component
- tests are passing


FrancescoCioria:
@veej rebased and fixed an import

PR #999, opened 2017-07-14
#901: backport mobile optimizations from alinity pro (closes #901)

### LINKED ISSUE #901: backport mobile optimizations from alinity pro
## requirements
on Alinity PRO we've added a mobile optimized version for:
- Tooltip
- TextOverflow
- (Popover?)

We should backport those wrappers here so to facilitate their use on other projects that support mobile/tablet devices

## specs
simply add the new file in the same folder and that's it :)

NOTE: for now don't update examples as showroom is not meant for mobile anyway...

## misc
{optional: other useful info}

ascariandrea:
Closes #901

### tests performed
- tested on *alinity-pro* ✅ 

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
     "react-transition-group": "1.1.1",
     "sass-flex-mixins": "^0.1.0",
     "tcomb-react": "^0.9.0",
-    "tlds": "^1.185.0"
+    "tlds": "^1.185.0",
+    "typelevel-ts": "^0.1.2"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -97,6 +97,7 @@
     "smooth-release": "^8.0.0",
     "style-loader": "^0.17.0",
     "stylelint-webpack-plugin": "^0.7.0",
+    "typelevel-ts": "^0.1.2",
```
ascariandrea: I've put it in `devDeps` cause it's like this in *alinity-pro*, but maybe in this case should be a `dep`

ascariandrea: Done! (let me rewrite the branch's history before merge)

PR #1004, opened 2017-07-19
Popover anchor and re render

ascariandrea:
To resolve this [issue](https://github.omnilab.our.buildo.io/buildo/alinity-pro/issues/6049) I came up with this solution.
Removing the `saveValuesFromNodeTree` in `componentWillReceiveProps` seems to have no cons.
Since you have more domain knowledge can you give me your feedback on this?
Thansk

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -53,6 +53,7 @@ export default class Popover extends React.Component {
   state = { isOpen: false };
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


ascariandrea:
The re-rerender was caused by the different instance of `ContextWrapper` returned by `getContextWrapper` utility method.
So, the old `ContextWrapper` component was unmounted while the new one was mounted each time the `appendPopover` or `componentDidUpdate` was called.

PR #1024, opened 2017-08-02
#1023: Popover: add logic to position it completely automatically (closes #1023)

### LINKED ISSUE #1023: Popover: add logic to position it completely automatically
## requirements
`anchor="auto"` is a good start, but we need the `Dropdown` to position itself completely automatically which means:
- choose position
- choose anchor

use `position` and `anchor` props as defaults in case there enough space for the `Popover`

## specs
for now don't move the arrow dynamically as it complicates the logic a lot. instead search for a combination of `postion`+`anchor` that works:
- start with default position and anchor
- if they're not ok
  - position ok; anchor not ok
    - search for anchor that works. if none -> change position and repeat
  - position not ok -> change position and repeat
- if no combination of position and anchor works -> log warning and use defaults

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1023

### tests performed
tested on AlinityPRO ->`Tooltip` is re-positioned correctly when there isn't enough space

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -508,7 +508,7 @@ export default class Popover extends React.Component {
     }
 
     return {
-      position: 'absolute',
+      position: isAbsolute ? 'fixed' : 'absolute',
       top: (isAbsolute ? child.y : 0) + (positionOffset.top + anchorOffset.top + offsetY),
       left: (isAbsolute ? child.x : 0) + (positionOffset.left + anchorOffset.left + offsetX),
       maxWidth
--- a/src/Popover/Popover.js
+++ b/src/Popover/Popover.js
@@ -14,7 +14,7 @@ export const Props = {
     auto: t.maybe(t.Boolean),
     attachToBody: t.maybe(t.Boolean),
     position: t.maybe(t.enums.of(['top', 'bottom', 'left', 'right'])),
-    anchor: t.maybe(t.enums.of(['start', 'center', 'end', 'auto'])),
+    anchor: t.maybe(t.enums.of(['start', 'center', 'end'])),
     event: t.maybe(t.enums.of(['click', 'hover'])),
     onShow: t.maybe(t.Function),
     onHide: t.maybe(t.Function),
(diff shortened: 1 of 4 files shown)


src/Popover/Popover.js
```diff
@@ -371,7 +376,50 @@ export default class Popover extends React.Component {
     );
   };
 
-  getVisiblePopover = () => this.popoverTemplate(this.computePopoverStyle());
+  getVisiblePopover = () => {
+    const { popover } = this.state;
+    const popoverProps = this.getPopoverProps();
+
+    if (popoverProps.auto) {
+      const positions = uniq([popoverProps.position, 'top', 'bottom', 'left', 'right']);
```
giogonzo: I don't understand this line.. you're listing all 4 possible positions anyway? :thinking: 
isn't this the same as `positions = ['top', 'bottom', 'left', 'right']`?

src/Popover/Popover.js
```diff
@@ -371,7 +376,50 @@ export default class Popover extends React.Component {
...
+
+      const popoverStyle = positions.reduce((acc, p) => {
+        if (acc === null) {
+          const anchors = uniq(
+            (p === popoverProps.position ? [popoverProps.anchor] : []).concat(['center', 'start', 'end'])
```
giogonzo: same comment here..

src/Popover/Popover.js
```diff
@@ -479,12 +470,25 @@ export default class Popover extends React.Component {
...
+        break;
+      case 'end':
+        anchorOffset.left = isHorizontal ? (child.width - popover.width) : 0;
+        anchorOffset.top = isVertical ? (child.height - popover.height) : 0;
+        break;
+    }
+
     return {
-      position: 'absolute',
+      position: isAbsolute ? 'fixed' : 'absolute',
```
giogonzo: why this change?

giogonzo:
@FrancescoCioria also, this is the first tooltip I tested on aliniq:
screen shot 2017-08-03 at 10 01 24 am
Is this branch outdated?

PR #1026, opened 2017-08-03
#1025: Implement `Overflow` (closes #1025)

### LINKED ISSUE #1025: Implement `ComponentOverflow`
## requirements
as we've done with `TextOverflow`, it would be useful to have a `ComponentOverflow` that checks if the component is not entirely displayed and, in this case, shows a custom placeholder on the right and renders the entire component in a `Popover` when hovered.

## specs
- move logic from `TextOverflow` to generic `Overflow` component
- use it to implement `ComponentOverflow`

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1025

### tests performed
tested a POC on AlinityPRO -> worked fine

tested with following example:
[https://gyazo.com/159b1f888d8fbca32144f415761ae9e2](https://gyazo.com/159b1f888d8fbca32144f415761ae9e2)

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/Overflow/Overflow.js
+++ b/src/Overflow/Overflow.js
@@ -25,6 +25,10 @@ export default class Overflow extends React.Component {
     this.verifyOverflow();
   }
 
+  componentDidUpdate() {
+    this.verifyOverflow();
+  }
+
   getElementWidth(element) {
     if (element && typeof window !== 'undefined') {
       return parseFloat(window.getComputedStyle(element).width);
--- a/src/Overflow/Overflow.js
+++ b/src/Overflow/Overflow.js
@@ -1,14 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import debounce from 'lodash/debounce';
 import { props, t } from '../utils';
 import ResizeSensor from '../ResizeSensor/ResizeSensor';
 
 export const Props = {
   children: t.ReactChild,
   onChange: t.Function,
-  verifyOverflowOn: t.maybe(t.enums.of(['resize', 'hover'])),
-  debounceVerifyOverflow: t.maybe(t.Number),
   id: t.maybe(t.String),
   className: t.maybe(t.String),
   style: t.maybe(t.Object)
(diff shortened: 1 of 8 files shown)


FrancescoCioria:
@giogonzo please don't merge it yet

src/Overflow/Overflow.js
```diff
@@ -0,0 +1,66 @@
...
+export const Props = {
+  children: t.ReactChild,
+  onChange: t.Function,
+  id: t.maybe(t.String),
+  className: t.maybe(t.String),
+  style: t.maybe(t.Object)
+};
+
+/**
+ * Util component which accepts calls a callback whenever the content starts or stop overflowing.
```
giogonzo: typo `accepts calls`

src/Overflow/Overflow.js
```diff
@@ -0,0 +1,66 @@
...
+ * @param children - a function that will be called with the argument
+ * @param onChange - tooltip delay if the component is lazy
```
giogonzo: outdated descriptions :D

src/Overflow/Overflow.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  onResize = () => this.verifyOverflow()
+
+  render() {
+    const { children, style, className, id } = this.props;
+
+    return (
+      <ResizeSensor debounce={10} onResize={this.onResize}>
```
giogonzo: magic number! Is it the same we are using in other components?

src/Overflow/README.md
```diff
@@ -0,0 +1,14 @@
+# Overflow
+
+Util component which accepts calls a callback whenever the content starts or stop overflowing.
+
+## Props
```
giogonzo: looks out of sync compared to the real props

src/Overflow/Overflow.js
ecamellini: starts or stop**s** (s missing)

src/Overflow/Overflow.js
```diff
@@ -0,0 +1,66 @@
...
+        <div {...{ className, id }} style={{ ...style, width: '100%' }} ref={ref => { this.ref = ref; }}>
```
ecamellini: Why `div` and not `FlexView`?

ecamellini: The wrapper (`div`, `FlexView` etc) could be passed as a prop maybe, like it happens in other BRCs

src/Overflow/README.md
```diff
@@ -0,0 +1,14 @@
...
+|Name|Type|Default|Description|
+|----|----|-------|-----------|
+| **children** | <code>ReactChildren</code> |  | **required**. A function that will be called with the argument |
+| **onChange** | <code>Function</code> |  | **required**. Tooltip delay if the component is lazy |
+| **verifyOverflowOn** | <code>enum("resize" &#124; "hover")</code> | <code>"resize"</code> | *optional*. Check if the content is overflowing on one of the following events: "resize" &#124; "hover" |
```
ecamellini: Not a prop

src/Overflow/README.md
```diff
@@ -0,0 +1,14 @@
...
+| **debounceVerifyOverflow** | <code>Number</code> |  | *optional*. Use this to debounce the calls to verify if the content is overflowing |
```
ecamellini: Not a prop

src/Overflow/README.md
```diff
@@ -0,0 +1,14 @@
+# Overflow
+
+Util component which accepts calls a callback whenever the content starts or stop overflowing.
+
+## Props
+|Name|Type|Default|Description|
+|----|----|-------|-----------|
+| **children** | <code>ReactChildren</code> |  | **required**. A function that will be called with the argument |
+| **onChange** | <code>Function</code> |  | **required**. Tooltip delay if the component is lazy |
```
ecamellini: "Tooltip delay if the component is lazy" and "A function that will be called with the argument" are not the right descriptions for these props

PR #1030, opened 2017-08-07
#1029: ScrollView: scrollbars are visible even when thumb is hidden (closes #1029)

### LINKED ISSUE #1029: ScrollView: scrollbars are visible even when thumb is hidden
## description
scrollbars are visible even when thumb is hidden and if you hover one of them the `:hover` style is activated

## how to reproduce

## specs
This is due to the `padding` of the two scrollbars and the fact that, through pure CSS, we can select only `thumb`s when the scrollbar is hidden.
We should use JS to add a class `visible` to each scrollbar if it's visible

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1029

### tests performed
tested on AlinityPRO:
- scrollbars work as before
- they're no longer visible/hoverable when thumb has height/width `0`
- if `children` use `PureComponent` (as we're doing) no useless render
  - tested on `UserPreferences` page: same number of renders for `UserRow`

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/ScrollView/ScrollView.js
+++ b/src/ScrollView/ScrollView.js
@@ -38,6 +38,11 @@ export default class ScrollView extends React.PureComponent {
     innerComponent: 'div'
   }
 
+  state = {
+    isVerticalScrollbarVisible: false,
+    isHorizontalScrollbarVisible: false
+  }
+
   /**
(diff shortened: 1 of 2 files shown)


francescogior:
@FrancescoCioria I rebased it on master

PR #1033, opened 2017-08-17
#1032: Image: copy-paste Image from buildo.io (closes #1032)

### LINKED ISSUE #1032: Image: copy-paste Image from buildo.io
## requirements
copy-paste `Image` from buildo.io

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1032

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/Image/getUrl.js
+++ b/src/Image/getUrl.js
@@ -1,9 +1,5 @@
 const cloudinaryId = 'buildo';
 const cloudinaryBaseUrl = `https://res.cloudinary.com/${cloudinaryId}/image/fetch/`;
-const defaultParams = {
-  quality: 'auto',
-  format: 'auto'
-};
 
 function isRetinaDisplay() {
   if (window.matchMedia) {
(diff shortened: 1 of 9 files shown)


src/Image/getUrl.js
```diff
@@ -0,0 +1,27 @@
...
+const cloudinaryBaseUrl = `https://res.cloudinary.com/${cloudinaryId}/image/fetch/`;
+
+function isRetinaDisplay() {
+  if (window.matchMedia) {
+    const mq = window.matchMedia('only screen and (min--moz-device-pixel-ratio: 1.3), only screen and (-o-min-device-pixel-ratio: 2.6/2), only screen and (-webkit-min-device-pixel-ratio: 1.3), only screen  and (min-device-pixel-ratio: 1.3), only screen and (min-resolution: 1.3dppx)');
+    return (mq && mq.matches || (window.devicePixelRatio > 1));
+  }
+}
+
+function getUrl(src, { width: _width, height: _height, quality: _quality, format: _format }) {
```
francescogior: I would add `= {}` to the second parameters since all the inner props are optional

francescogior:
@FrancescoCioria also change the example image :) 
image

src/Image/getUrl.js
FrancescoCioria: 👍

PR #1039, opened 2017-08-18
#1035: [Typo] Initial setup (closes #1035)

### LINKED ISSUE #1035: [Typo] Initial setup
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

veej:
Closes #1035

### tests performed
- [x] showroom works
- [x] build works
- [x] AlinityPRO works with new build

and all three keep working from #1041 branch! (so they work when `src` has one or more `.tsx?` files)

- [ ] Google Chrome
- [ ] Internet Explorer


FrancescoCioria:
@veej It finally works! the commit history is horrible, but who cares...
in the end I made it work by making both `typescript` and `babel` transpile every file both in the build and in webpack... it's not a great solution, I know, but it works :P

UPDATE: ups, the tests are still failing...

ascariandrea:
Shouldn't the lodash imports case be fixed also in `package.json`?

FrancescoCioria:
> Shouldn't the lodash imports case be fixed also in package.json?

what do you mean? they are correct in the `package.json`: they should all be in lowercase

ascariandrea:
It's confusing the different case in `import` and *dependencies* in `package.json`.
Won't the `import` fail for case sensitive os?

FrancescoCioria:
> It's confusing the different case in import and dependencies in package.json.
Won't the import fail for case sensitive os?

can you make an example? I really don't understand: I fixed every case error of `lodash` and the CI (which is case sensitive) is passing. Which code are you reading?

ascariandrea:
My bad 🙈 
I was looking at the previous case instead the fixed one 😅

PR #1041, opened 2017-08-18
#1040: [Typo] AsyncStatusIndicator (closes #1040)

### LINKED ISSUE #1040: [Typo] AsyncStatusIndicator
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #1040

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/AsyncStatusIndicator/AsyncStatusIndicator.tsx
```diff
@@ -27,16 +52,16 @@ export const Props = {
  */
 @skinnable()
 @props(Props)
-export default class AsyncStatusIndicator extends React.PureComponent {
+export default class AsyncStatusIndicator extends React.PureComponent<AsyncStatusIndicatorProps, void> {
```
gabro: you don't need `void`. `React.PureComponent<AsyncStatusIndicatorProps>` works the same

src/AsyncStatusIndicator/AsyncStatusIndicator.tsx
```diff
@@ -27,16 +52,16 @@ export const Props = {
  */
 @skinnable()
```
FrancescoCioria: I would remove @skinnable completely from brc as we've been doing in AlinityPRO and GDSM

src/AsyncStatusIndicator/AsyncStatusIndicator.d.ts
```diff
@@ -1,20 +1,6 @@
 import { CSSProperties, PureComponent } from 'react';
```
FrancescoCioria: this file should be completely removed: Typescript will automatically generate it

FrancescoCioria:
@ascariandrea I further improved the build in #1039 to:
- transpile `.tsx?` with both `tsc` and `babel`
- generate `.d.ts` automatically for `.tsx?`
- keep using only `babel` for `.js`
- keep moving manually written `.d.ts` from `src` to `lib`

now we can completely remove `.d.ts` and write `.tsx?` files as we would if we didn't have `babel` at all.
I rebased it on `1035-...` and added some commits.

NOTE: the config `declaration: true` requires `allowJs` to be `false` -> I had to refactor `utils` in TypeScript to make it work.

gabro:
Since VSCode will look for docs on the props, can we also move them inline?

I mean

```ts
type Props = {
  /** the class name */
  className?: string
}

/** The Foo */
export default class Foo extends React.Component<Props>
```

instead of 

```ts
/** The Foo
  * @param className the class name
  */
export default class Foo extends React.Component<Props>
```

src/AsyncStatusIndicator/AsyncStatusIndicator.tsx
giogonzo: also don't use void, it should be `State extends {}`.
`void` used to break in weird ways with `@types/react@15.0.38`

giogonzo: (probably not anymore in 16 :D https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L269)

gabro:
rebased

PR #1054, opened 2017-08-23
#1052: [Typo] Browser Detector (closes #1052)

### LINKED ISSUE #1052: [Typo] Browser Detector
← #1034

## specs
use `bowser` typings

gabro:
Closes #1052

### tests performed
unit tests


src/BrowserDetector/BrowserDetector.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import some = require('lodash/some');
+import { props, t, ReactChildren } from '../utils';
+import * as bowser from 'bowser';
+
+// FIXME: this is hardcoded and there's no guarantee it matches with bowser's types
+const browsers = [
```
gabro: I've copied over all the browsers from the bowser typings.

test/components/BrowserDetector.test.tsx
```diff
@@ -1,10 +1,10 @@
-import React from 'react';
-import renderer from 'react-test-renderer';
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
 
 import BrowserDetector from '../../src/BrowserDetector';
 
 function forceUserAgent(ua) {
-  navigator.__defineGetter__('userAgent', () => ua);
+  navigator['__defineGetter__']('userAgent', () => ua);
```
gabro: TS doesn't know about `__defineGetter__`

src/BrowserDetector/BrowserDetector.tsx
giogonzo: Y U NO `@types/bowser`?

gabro: these are values.

Also, no need for `@types/bowser`, since https://github.com/lancedikson/bowser/blob/master/typings.d.ts

test/components/BrowserDetector.test.tsx
gabro: @giogonzo `__defineGetter__` is this one https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__

In theory it could be replaced by

```ts
Object.defineProperty(navigator, 'userAgent', { get: () => ua })
```

but it throws an error when calling it multiple times, like I'm doing in these tests. One possible solution is something like


```ts
let _ua = '';
Object.defineProperty(navigator, 'userAgent', { get: () => _ua })
```

and then just change the `_ua` value, but I didn't like it and `__defineGetter__` works as expected, so... ¯\_(ツ)_/¯

PR #1055, opened 2017-08-23
#1053: [Typo] Button (closes #1053)

### LINKED ISSUE #1053: [Typo] Button
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #1053

### tests performed
- test pass ✅ 
- works on alinity-pro ✅ 

During the refactor, I've noticed some that all classes, events, style and so on are assigned to `.button-inner` and I wondered why.
If I pass a `className` prop to a component I expect to be added to the top level element.

What was the reason behind this strategy?

- [ ] Google Chrome
- [ ] Internet Explorer


src/Button/Button.tsx
```diff
@@ -0,0 +1,227 @@
...
+  children?: string,
+  /** type of the button (default, primary, positive, negative, flat) */
+  type?: ButtonProps.ButtonType,
+  /** shortcut for type "primary" */
+  primary?: boolean,
+  /** shortcut for type "flat" */
+  flat?: boolean,
+  /** size of the button, one of 'tiny', 'small', 'medium' */
+  size?: ButtonProps.ButtonSize,
+  /** fluid - fluid (block) button, takes the width of the container */
```
gabro: remove `fluid -`

src/Button/Button.tsx
```diff
@@ -0,0 +1,227 @@
...
+  fluid?: boolean,
+  /** circular - circular button, this is allowed only if it's an icon button */
```
gabro: remove `circular - `

src/Button/Button.tsx
```diff
@@ -0,0 +1,227 @@
...
+  circular?: boolean,
+  /** textOverflow - function to handle the overflow of too long labels, replacing with ellipsed string and tooltip */
```
gabro: remove `textOverflow - `

src/Button/Button.tsx
```diff
@@ -0,0 +1,227 @@
...
+  textOverflow?: TextOverflowCompatibleComponent,
+  style?: React.CSSProperties,
```
gabro: you can copy the description of `style`/`className` from the same props in `AsyncStatusIndicator`

ascariandrea: The description from `AsyncStatusIndicator` is not applicable to `Button` component as it says:
> /** an optional style object to pass to top level element of the component */

and, as I pointed out in the PR description, the `style` is assigned to the `.button-inner` element.
This is not a big deal, but it's incorrect anyway.

I'll rephrase as:
> /** an optional style object to pass to first inner element of the component */

is it better?

gabro: sure!

ascariandrea:
@gabro should be ready to review!

package.json
```diff
@@ -41,6 +41,7 @@
   "homepage": "https://github.com/buildo/react-components",
   "dependencies": {
     "@types/lodash": "4.14.74",
+    "@types/node": "^8.0.24",
```
FrancescoCioria: why? BRC is a browser lib, why should we need these?
if they're for tests please move to dev-deps

also, **remove the ^**!!1!11!!

gabro: I think this is for `process.env`

gabro: but I wouldn't use `@types/node` any way since this is not node, but `process.env` is inserted by webpack. I would just follow the approach suggested here: https://medium.com/the-missing-bit/using-node-env-with-typescript-and-webpack-689a85041978

gabro: but using `const` instead of `var` :P

src/Button/Button.tsx
```diff
@@ -0,0 +1,225 @@
...
+  icon: string | ButtonStateMap;
+  /** an optional class name to pass to first inner element of the component */
+  className: string;
+  /** an optional style object to pass to first inner element of the component */
+  style: object
+};
+
+
+export type ButtonProps = ButtonRequiredProps & Partial<ButtonDefaultProps>;
+export type ButtonPropTypes = ButtonProps;
```
gabro: What is this?

src/Button/StatefulButton.tsx
```diff
@@ -82,9 +113,9 @@ export default class StatefulButton extends React.PureComponent {
     }
   }
 
-  componentWillReceiveProps(props) {
-    if (process.env.NODE_ENV !== 'production') {
-      t.assert(props.stableSuccess === this.props.stableSuccess);
+  componentWillReceiveProps(props: StatefulButtonProps) {
+    if ((process as NodeJS.Process).env.NODE_ENV !== 'production') {
```
gabro: I would change this as discussed above.

package.json
FrancescoCioria: I `declare`d it in `utils/log.ts`
we can `declare` it globally

ascariandrea: @FrancescoCioria you shoul remove the `^` too: https://github.com/buildo/react-components/blame/master/package.json#L74 !11!!!!!

src/Button/Button.tsx
ascariandrea: It was exported in the `.d.ts` before, so I've just reported here. It was probably used in alinity-pro, but we can remove it, and then fix in the other projects (if needed)

ascariandrea: @gabro Should I delete it then?

gabro: I would remove it, yes.

typings/typings.d.ts
```diff
@@ -0,0 +1 @@
+export declare const process : { env: { NODE_ENV: string } }
```
gabro: extra space after `process` 🤓

gabro: Also, I would declare this as 

```ts
export declare const process: { env: { NODE_ENV: 'development' | 'production' } };
```

so that this is a type error:

```ts
if (process.env.NODE_ENV !== 'productio') { // <-- typo
  // ...
}
```

src/Button/Button.tsx
ascariandrea: Removed :)

typings/typings.d.ts
ascariandrea: Type fixed!

src/TextOverflow/TextOverflow.js
```diff
@@ -9,7 +9,11 @@ import ResizeSensor from '../ResizeSensor/ResizeSensor';
 
 export const Props = {
   children: t.maybe(t.Function),
-  label: t.String,
+  label: t.maybe(t.union([t.String, t.Number])),
+  popover: t.maybe(t.interface({
```
FrancescoCioria: can you add below a description for the new `popover` prop?

ascariandrea: Done 👍

PR #1061, opened 2017-08-25
#1059: [Typo] ConfirmationInput (closes #1059)

### LINKED ISSUE #1059: [Typo] ConfirmationInput
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

veej:
Closes #1059

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


typings/react-input-children.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-input-children';
```
veej: 😢

giogonzo:
@gabro rationale? I get the idea, but does it actually make writing the component's internal simpler/cleaner? Otherwise is it just a matter of style (could add clarity I think)?

veej:
@gabro better now?

src/ConfirmationInput/ConfirmationInput.tsx
```diff
@@ -0,0 +1,217 @@
...
+  text?: string,
+  icon?: string,
+  onMouseDown: () => void,
+  onMouseEnter: () => void,
+  onMouseLeave: () => void
+} | undefined
+
+@props(Props, { strict: false })
+export default class ConfirmationInput extends React.PureComponent<ConfirmationInputProps, ConfirmationInputState> {
+
```
FrancescoCioria: @veej we should temporarily re-add the `defaultProps` here to avoid breaking `@props` validation.

(like @gabro did in this commit: https://github.com/buildo/react-components/pull/1077/commits/f6dd3aa00f448bc7138507624242d95f72fdeb38)

FrancescoCioria: nevermind, we'll do it in a separate issue

PR #1074, opened 2017-08-25
#1064: [Typo] DatePicker (closes #1064)

### LINKED ISSUE #1064: [Typo] DatePicker
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

veej:
Closes #1064

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,241 @@
...
+export type Value = string | Date | moment.Moment;
+
+export type DatePickerProps = {
+  /** an optional id to pass to top level element of the component */
+  id?: string,
+  /** current date */
+  value?: Value,
+  /** default date */
+  defaultValue?: Value,
+  //** called when value changes */
```
giogonzo: double `//`?

src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,241 @@
...
+import cx from '../utils/classnames';
+import { SingleDatePicker } from 'react-dates';
+import FlexView from 'react-flexview';
+import { props, t } from '../utils';
+import * as moment from 'moment';
+import Icon from '../Icon';
+
+export type Value = string | Date | moment.Moment;
+
+export type DatePickerProps = {
```
giogonzo: suggestion: doesn't affect JS users, affects TS users only (= 0)
what about a type parameter for the value type here?

```ts
type DatePickerProps<V extends Value> = {
  value?: V
  defaultValue?: V
  onChange?: (date?: V) => void
  // more props
}
// and
export default class DatePicker<V extends Value> extends React.PureComponent<Props<V>, State> { ... }
```

since inChange param type could be affected by `returnFormat` props, I guess:
```ts
type OnChangeProps<V extends Value> = {
  returnFormat?: never
  onChange?: (date?: V) => void
} | {
  returnFormat: string
  onChange?: (date?: string) => void
}
type DatePickerProps<V extends Value> = OnChangeProps<V> & {
  value?: V
  defaultValue?: V
  // more props
}
```
usage from TS would then be (in order to obtain stricter types for props)
```ts
class MyDatePicker extends DatePicker<string> // or moment.Date, etc.
// then use MyDatePicker normally
```
bonus: forbid using it without first having specified the precise value type: add `never` as default type parameter
```ts
export default class DatePicker<V extends Value = never> extends React.PureComponent<Props<V>, State> { ... }
```

gabro: Big 👍 on this, but the I think the never default would make this breaking

giogonzo: why? Did a quick test with:
```ts
type Props<V extends {}> = { value?: V };
class DatePicker<V extends string = never> extends React.PureComponent<Props<V>, {}> {}
ReactDOM.render(<DatePicker />, document.documentElement);
ReactDOM.render(<DatePicker value='asdas' />, document.documentElement);  // Type '"asdas"' is not assignable to type 'undefined'
class MyDatePicker extends DatePicker<string> {}
ReactDOM.render(<MyDatePicker value='asdas' />, document.documentElement);
```
and seems to work as expected

veej: I think he's saying the issue would be `breaking`, for this reason:
> // Type '"asdas"' is not assignable to type 'undefined'

giogonzo: ah ops, sorry, ok got it now, but breaking for who? no one is using this from TS :P

PR #1075, opened 2017-08-25
#1065: [Typo] Divider (closes #1065)

### LINKED ISSUE #1065: [Typo] Divider
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

veej:
Closes #1065

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/Divider/Divider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  size: 'small',
+  style: {}
+};
+
+/**
+ * A simple component used to visually divide UI elements
+ */
+@props(Props)
+export default class Divider extends React.PureComponent<DividerProps> {
+
```
FrancescoCioria: we should re-add `defaultProps` to make `@props` happy
(as done in this commit: https://github.com/buildo/react-components/pull/1077/commits/f6dd3aa00f448bc7138507624242d95f72fdeb38)

PR #1084, opened 2017-08-28
Make all TS components conform to #1082

### LINKED ISSUE #1082: The hitchhiker's guide to default props with TypeScript
# 1. Meet the types!

Every component with defaulted props defines 4 type aliases. For example, given a component named `Foo`

```ts
export type FooDefaultProps = { ... };
export type FooRequiredProps = { ... };
export type FooProps = FooRequiredProps & Partial<FooDefaultProps>;
type FooDefaultedProps = FooRequiredProps & FooDefaultProps; // Not exported
```

Guidelines:
  - all props in `FooDefaultProps` must be required
     - i.e. `{ label: string }` and **NOT** `{ label?: string }` 
  - when a props in `FooRequiredProps` is optional, **double check it shouldn't have a default instead**

> **NOTE**
> names of these types will change slighly when we'll introduce namespaces, see #1083 

# 2. Use `static defaultProps`

Define default props values as a static on the component class and **make sure to explicitly annotate the type** (so you can't forget a default).

E.g.

```tsx
export default class Foo extends React.Component<FooProps> {
  static defaultProps: FooDefaultProps = {
    label: ''
  };
  // ...
}
```

**Why?** Because we want React to keep defaulting our props.

# 3. Cast when accessing

If you're accessing a props that has a default, apply a cast to `this.props`.

E.g.

```tsx
render() {
  const { label } = this.props as FooDefaultedProps;
  return <div>{label}</div>;
}
```

**Why?** `this.props` is already of the correct type, because React defaulted the props for us, but TS doesn't know. TypeScript will enforce this 🎉 

# 4. If you define a helper, call it `defaultedProps()`

If you find yourself casting a lot, define a helper property called `defaultedProps()` that does the casting once.

E.g.

```tsx
defaultedProps = () => this.props as FooDefaultedProps;

onChange = (value) => {
  const { onWhatever } = this.defaultedProps();
  onWhatever(value);
};

render() {
  const { label } = this.defaultedProps();
  return <div>{label}</div>;
}
```

gabro:
Closes #1082

### DIFF
--- a/src/BackgroundDimmer/BackgroundDimmer.tsx
+++ b/src/BackgroundDimmer/BackgroundDimmer.tsx
@@ -5,27 +5,11 @@ import { props, t, ReactChildren } from '../utils';
 import FlexView, { IProps as FlexViewProps } from 'react-flexview';
 import omit = require('lodash/omit');
 
-export type BackgroundDimmerProps = {
+export type BackgroundDimmerRequiredProps = {
   /** children nodes/elements */
   children: any, // TODO: t.ReactChildren
-  /** background-color */
-  color?: string,
-  /** opacity */
-  alpha?: number,
-  /** z-index (BackgroundDimmer has `position: fixed`) */
-  zIndex?: number,
-  /** avoid propagation for scroll events */
-  stopScrollPropagation?: boolean,
   /** called when user clicks outside children */
   onClickOutside?: (e: React.SyntheticEvent<HTMLDivElement>) => void,
-  /** centeredContentWrapper width */
-  width?: string | number,
-  /** centeredContentWrapper max-width */
-  maxWidth?: string | number,
-  /** centeredContentWrapper height */
-  height?: string | number,
-  /** centeredContentWrapper max-height */
-  maxHeight?: string | number,
   /** component's class */
   className?: string,
   /** component's id */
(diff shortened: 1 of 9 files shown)


src/ConfirmationInput/ConfirmationInput.tsx
```diff
@@ -40,7 +40,8 @@ export type ConfirmationInputDefaultProps = {
   onClear: () => void,
 };
 
-export type ConfirmationInputProps = ConfirmationInputRequiredProps & Partial<ConfirmationInputDefaultProps>
+export type ConfirmationInputProps = ConfirmationInputRequiredProps & Partial<ConfirmationInputDefaultProps>;
+export type ConfirmationInputDefaultedProps = ConfirmationInputRequiredProps & ConfirmationInputDefaultProps;
```
veej: no need to export this

src/BackgroundDimmer/BackgroundDimmer.tsx
```diff
@@ -35,14 +19,26 @@ export type BackgroundDimmerProps = {
...
+  /** centeredContentWrapper max-width */
   maxWidth: string | number,
+  /** centeredContentWrapper height */
   height: string | number,
-  maxHeight: string | number
-}
+  /** centeredContentWrapper max-height */
+  maxHeight: string | number,
+  /** avoid propagation for scroll events */
+  stopScrollPropagation: boolean,
```
gabro: This was implicitly `undefined`, I explicitly made it `false` as a default prop.

src/Divider/Divider.tsx
```diff
@@ -17,6 +17,7 @@ export type DividerDefaultProps = {
 };
 
 export type DividerProps = Partial<DividerDefaultProps>;
+export type DividerDefaultedProps = DividerDefaultProps;
```
veej: no need to export this

src/BackgroundDimmer/BackgroundDimmer.tsx
```diff
@@ -134,7 +129,7 @@ export default class BackgroundDimmer extends React.PureComponent<BackgroundDimm
     };
 
     return (
-      <div className={cx('background-dimmer', className)} {...omit(props, ['onClickOutside'])}>
+      <div className={cx('background-dimmer', className)} {...omit(props, ['onClickOutside', 'stopScrollPropagation'])}>
```
gabro: we were not filtering `stopScrollPropagation` (which is not a valid property for `div`).

The tests didn't raise a warning because we were implicitly passing `undefined`, now that we're passing `false` it became evident. I'm not sure whether to consider this as a breaking change (it removes a React warning).

src/Divider/Divider.tsx
gabro: purely for naming consistency in the component implementation.

gabro:
@veej thanks, fixed!

src/BackgroundDimmer/BackgroundDimmer.tsx
gabro: @FrancescoCioria you're the `breaking` guy, what do you think?

FrancescoCioria: looks like a simple bug/defect so no breaking IMO

PR #1087, opened 2017-08-29
#1070: [Typo] Image (closes #1070)

### LINKED ISSUE #1070: [Typo] Image
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

ascariandrea:
Closes #1070

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/Image/Image.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const Props = {
+  src: t.String,
+  width: t.maybe(t.Number),
+  height: t.maybe(t.Number),
+  quality: t.maybe(t.union([t.Number, t.enums.of(['auto'])])),
+  format: t.maybe(t.enums.of(['jpeg', 'png', 'wdp', 'gif', 'auto']))
+};
+
+/** A replacement for `<img>` to serve optimized images through a CDN in production */
+@props(Props, { strict: false })
```
veej: Since we're using `strict: false`, the component should be able to receive every prop you can pass to `` too. 
I would add `& React.ImgHTMLAttributes<HTMLImageElement` to `ImageProps`

ascariandrea: Good catch, done!

veej: could you please fix exports too?

ascariandrea: Which one?

veej: see the general comment to this PR. you're not exporting `Props` and `ImageProps` from `index.ts`

ascariandrea: Sorry, I missed that line!

src/Image/Image.tsx
```diff
@@ -0,0 +1,39 @@
...
+export default class Image extends React.Component<ImageProps> {
```
veej: whoops, missed this. 
This should be a `PureComponent`

ascariandrea: But it wasn't before :(

veej: ok, leave it as is, even if it shouldn't hurt

src/Image/index.ts
```diff
@@ -0,0 +1,5 @@
+import Image, { ImageProps, Props } from './Image';
```
FrancescoCioria: why are we exporting tcomb `Props` from here? we're not doing in the other components

ascariandrea: Just addressed the @veej's comment, let me remove it

ascariandrea: I've amended the last commit and should be fine now!
Anyway, there are other components that export `Props` from `index.ts`, so we need to fix them too

veej: whoops... it seems only `BackgroundDimmer` is exporting them.

ascariandrea: And `ConfirmationInput`

PR #1090, opened 2017-08-29
#1089: [Typo] Popover (closes #1089)

### LINKED ISSUE #1089: [Typo] Popover
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

gabro:
Closes #1089

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/Popover/Popover.tsx
```diff
@@ -145,13 +215,13 @@ export default class Popover extends React.Component {
...
   // extend with default values
-  getPopoverProps = (_props) => {
+  getPopoverProps = (_props?: PopoverProps) => {
     const props = _props || this.props;
     return {
-      type: 'relative',
-      position: 'top',
-      anchor: 'center',
-      event: 'hover',
+      // type: 'relative', //FIXME(gabro): doesn't seem used
```
gabro: double check, it wasn't defined in the props in the `.d.ts` files and I didn't find any usage.

src/Popover/Popover.tsx
```diff
@@ -165,18 +235,18 @@ export default class Popover extends React.Component {
     };
   };
 
-  getPopoverNode = () => {
-    let popover;
+  getPopoverNode = (): Element => {
+    let popover: Element;
     if (this.isAbsolute()) {
-      popover = this.popoverNode;
+      popover = this.popoverNode!; // FIXME(gabro): is this safe?
```
gabro: this doesn't change the current behavior, so it's probably safe, still I would like to double check to see whether we have potential bugs.

src/Popover/Popover.tsx
```diff
@@ -289,21 +359,22 @@ export default class Popover extends React.Component {
...
-
-  eventWrapper = cb => e => {
+  eventWrapper = <E extends React.SyntheticEvent<HTMLElement>>(cb: (e: E) => void) => (e: E) => {
     const { event } = this.getPopoverProps();
-    const childrenNode = ReactDOM.findDOMNode(this.refs.children).childNodes[0];
-    const el = e.target || e.srcElement;
+    const childrenNode = ReactDOM.findDOMNode(this.refs.children).children[0];
+    // It's safe to assume that the target is going to be a DOM Element
+    // See also: https://stackoverflow.com/questions/28900077/why-is-event-target-not-element-in-typescript
+    const el = e.target as Element;
```
gabro: see the comment, in general it seems harmless.

src/Popover/Popover.tsx
```diff
@@ -289,21 +359,22 @@ export default class Popover extends React.Component {
...
     if (this.isAbsolute() || (event === 'hover') || this.isEventInsideTarget(el, childrenNode)) {
       cb(e);
     }
   };
 
-  getDelayWhenClosed = delay => (delay || {}).whenClosed || delay;
+  getDelayWhenClosed = (delay?: PopoverProps.Delay) => (typeof delay === 'number') ? delay : (delay || {}).whenClosed;
```
gabro: delay can be number or an object, so TS explicitly forces us to check.

src/Popover/Popover.tsx
```diff
@@ -321,23 +392,23 @@ export default class Popover extends React.Component {
...
 
-  onMouseEvent = ({ type }) => {
+  onMouseEvent: React.MouseEventHandler<any> = ({ type }) => {
     const { delay } = this.getPopoverProps();
     const delayWhenClosed = this.getDelayWhenClosed(delay);
     const delayWhenOpen = this.getDelayWhenOpen(delay);
 
     if (this.isOpen()) {
-      return delayWhenOpen ? this.onMouseEventDebouncedWhenOpen(type) : this._onMouseEvent(type);
+      return delayWhenOpen ? this.onMouseEventDebouncedWhenOpen && this.onMouseEventDebouncedWhenOpen(type) : this._onMouseEvent(type);
```
gabro: very naive way of handling this, but I don't have better options. I didn't find any guarantee that `onMouseEventDebouncedWhenOpen` is never null.

src/Popover/Popover.tsx
```diff
@@ -321,23 +392,23 @@ export default class Popover extends React.Component {
...
     } else {
-      return delayWhenClosed ? this.onMouseEventDebouncedWhenClosed(type) : this._onMouseEvent(type);
+      return delayWhenClosed ? this.onMouseEventDebouncedWhenClosed && this.onMouseEventDebouncedWhenClosed(type) : this._onMouseEvent(type);
```
gabro: same as above

src/Popover/Popover.tsx
```diff
@@ -377,45 +448,45 @@ export default class Popover extends React.Component {
   };
 
   getVisiblePopover = () => {
-    const { popover } = this.state;
+    const popover = this.state.popover!; // TODO(gabro): is this safe?
```
gabro: this doesn't change the current behavior, so it's probably safe, still I would like to double check to see whether we have potential bugs.

src/Popover/Popover.tsx
```diff
@@ -377,45 +448,45 @@ export default class Popover extends React.Component {
...
-            if (workingPopoverStyle === null) {
-              const _popoverStyle = this.computePopoverStyle(p, a);
-
-              // not enough space
-              if (
-                _popoverStyle.top < 0 || _popoverStyle.top + popover.height > window.innerHeight ||
-                _popoverStyle.left < 0 || _popoverStyle.left + popover.width > window.innerWidth
-              ) {
-                return null;
+      const positions = uniq<PopoverProps.Position>([popoverProps.position, 'top', 'bottom', 'left', 'right']);
```
gabro: for some reason `uniq` didn't infer the type correctly.

src/Popover/Popover.tsx
```diff
@@ -446,8 +517,9 @@ export default class Popover extends React.Component {
     };
   };
 
-  computePopoverStyle = (position, anchor) => {
-    const { child, popover } = this.state;
+  computePopoverStyle = (position: PopoverProps.Position, anchor: PopoverProps.Anchor): PopoverStyle => {
+    const child = this.state.child!; // TODO(gabro): is this safe?
+    const popover = this.state.popover!; // TODO(gabro): is this safe?
```
gabro: same as the similar cases above.

src/Popover/Popover.tsx
```diff
@@ -498,13 +570,14 @@ export default class Popover extends React.Component {
     const isRelative = !this.isAbsolute();
     const isOpen = this.isOpen();
     const popover = isRelative && (this.initialized && isOpen ? this.getVisiblePopover() : this.getHiddenPopover());
+    const style: React.CSSProperties = {
```
gabro: moved style above so that we can enforce its type.

src/Popover/Popover.tsx
giogonzo: It's probably the same old story with string literal unions: `[popoverProps.position, 'top', 'bottom', 'left', 'right']` is inferred as `(Position | string)[]`

src/Popover/Popover.tsx
```diff
@@ -109,10 +177,12 @@ export default class Popover extends React.Component {
     }
   };
 
-  onClickOutside = (e) => {
+  onClickOutside = (e: MouseEvent) => {
     const childrenNode = ReactDOM.findDOMNode(this.refs.children);
-    const popoverNode = this.isAbsolute() ? this.containerNode : childrenNode.childNodes[1];
-    const el = e.target || e.srcElement;
+    const popoverNode = this.isAbsolute() ? this.containerNode : childrenNode.children[1];
```
gabro: I've changed every use of `childNodes` to `children` (won't be repeating this comment).

As a reminder: all elements are nodes, but not all nodes are elements.

This is potentially breaking, but we're accessing a lot of properties of `Element`, so this would break if `childNodes` ever returned a children. I think in practice we never access a node that is not an element, so it's more correct now.

src/Popover/Popover.tsx
FrancescoCioria: this can be removed: the prop `type` has been replaced by `attachToBody`

src/Popover/Popover.tsx
```diff
@@ -194,7 +264,7 @@ export default class Popover extends React.Component {
     return { top, left };
   };
 
-  saveValuesFromNodeTree = (cb) => {
+  saveValuesFromNodeTree = (cb?: () => any) => {
```
FrancescoCioria: actually `cb` is never passed to `saveValuesFromNodeTree` (it was until some months ago)...
I vote for removing it completely

src/Popover/Popover.tsx
FrancescoCioria: it's enforced by `updateDebouncedMousedEvents`  in `componentWillReceiveProps`
anyway, no problem in double checking here

(this flow is complex and confusing, I know)

src/Popover/Popover.tsx
FrancescoCioria: this is how it works:
1. is `this.initialized` is `false` we render `getHiddenPopover` which doesn't use any state variable
2. `state.popover` and `state.child` are created/updated by `saveValuesFromNodeTree` in `componentDidMount` and `componentWillReceiveProps` (so they should exist after the first `render`)
3. in `componentDidMount` we set `this.initialized` to `true`
4. from here `state.popover` and `state.child` should exist and can be used by `getVisiblePopover`

there may be some edge cases where this process fail, but we haven't had a bug since a lot of time. Still, i'm open to improvements :)

gabro: ok! I'll remove the comments

PR #1107, opened 2017-09-01
#1100: [Typo] Overflow (closes #1100)

### LINKED ISSUE #1100: [Typo] Overflow
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4471039)

← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1100

### tests performed
`npm start` -> it works without errors in console

- [ ] Google Chrome
- [ ] Internet Explorer


src/Overflow/Overflow.tsx
```diff
@@ -1,26 +1,38 @@
...
+};
+
+export type State = {
+  isOverflowing: boolean
+};
+
 export const Props = {
-  content: ReactNode,
-  contentIfOverflowing: ReactNode,
+  content: ReactElement,
```
FrancescoCioria: `Overflow` never supported a content that was a `string` or a `number` so I explicitly made it stricter here and in the TS type

FrancescoCioria:
@gabro perfect! you can go ahead and merge it if you want

PR #1109, opened 2017-09-01
#1108: [Typo] Toaster (closes #1108)

### LINKED ISSUE #1108: [Typo] Toaster
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4473527)

← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

gabro:
Closes #1108 

### tests performed
Still working in the showroom.


package.json
```diff
@@ -73,6 +73,7 @@
     "@types/enzyme": "2.8.6",
     "@types/jest": "20.0.7",
     "@types/node": "8.0.25",
+    "@types/react-transition-group": "2.0.4",
```
gabro: no 🎩

src/Toaster/TimerToast.tsx
```diff
@@ -12,21 +12,38 @@ export const Props = {
...
 
-  componentWillUnmount = () => this.clearTimer();
+  componentWillUnmount() {
+    this.clearTimer();
+  }
 
   resetTimer = () => {
     if (!this.timer && !this.state.completed) {
-      this.timer = setTimeout(this.onTimeout, this.props.duration);
+      this.timer = window.setTimeout(this.onTimeout, this.props.duration);
```
gabro: this `window` is just for helping VSCode picking up the right overload (otherwise it thinks it's node's `setTimeout` which returns a `Timer` instead of a `number`.

src/Toaster/Toaster.tsx
```diff
@@ -43,7 +64,7 @@ export default class Toaster extends React.Component {
   }
 
   componentDidMount() {
-    const node = this.props.attachTo ? this.toaster : ReactDOM.findDOMNode(this).parentNode;
+    const node = this.props.attachTo ? this.toaster! : ReactDOM.findDOMNode(this).parentElement!;
```
gabro: both `!` are safe: if `this.props.attachTo` is `true`, then `toaster` has been initialized in `componentWillMount` and `parentElement` is known be not null.

src/Toaster/Toaster.tsx
```diff
@@ -71,12 +92,13 @@ export default class Toaster extends React.Component {
...
     const { children, transitionStyles, transitionEnterTimeout, transitionLeaveTimeout } = this.props;
-    return React.Children.map(children, (el, i) => {
+    return React.Children.map(children, (_el, i) => {
+      const el = _el as React.ReactElement<any>;
       return (
         <TransitionWrapper
           {...{ transitionStyles, transitionEnterTimeout, transitionLeaveTimeout }}
           style={this.getTranslationStyle(i)}
-          key={el.key}
+          key={el.key || undefined}
```
gabro: this is stupid, but `TransitionWrapper` accepts `string | undefined`, while `el.key` is `string | null`...

src/utils/log.ts
```diff
@@ -1,4 +1,4 @@
-export function warn(content: string | (() => void)) {
+export function warn(content: any | (() => void)) {
```
gabro: we're calling `warn` with an array of `string | Element` above, so I just put `any` here. It shouldn't cause problems since `console.warn` takes `any`...

gabro:
@FrancescoCioria done!

PR #1112, opened 2017-09-01
#1103: [Typo] Navbar (closes #1103)

### LINKED ISSUE #1103: [Typo] Navbar
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4472413)

← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

francescogior:
Closes #1103

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/NavBar/index.d.ts
```diff
@@ -1,3 +0,0 @@
-import NavBar, { NavBarProps } from './NavBar';
-export default NavBar;
-export { NavBarProps };
```
FrancescoCioria: @francescogior please add the export of `NavBarProps` to `index.ts`

src/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { props, t, ReactChildren } from '../utils';
+import FlexView from 'react-flexview';
+
+export type NavBarProps = {
+  /** Left content: It doesn't shrink nor grow */
+  /** Center content. It grows to use all the available space */
+  /** Right content. It doesn't shrink nor grow */
+  /** Max Width. For css `max-width` */
+  content: {
+    left?: any, // TODO: ReactChildren
```
veej: `any`? couldn't be a `JSX.Element | string`?
(same for every other `any` below)

src/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,84 @@
...
+    } = this.props;
+    const style = {
+      ..._style,
+      position: fixed ? 'fixed' : undefined,
+      top: fixed ? 0 : undefined,
+      zIndex: fixed ? 99999 : undefined,
+      width: '100%',
+      height,
+      background
+    } as React.CSSProperties;
```
veej: why the conversion here? Should't be better to define `const style: React.CSSProperties = { ... }`?

src/NavBar/index.ts
```diff
@@ -1,2 +1,3 @@
 import NavBar from './NavBar';
 export default NavBar;
+export { NavBarProps } from './NavBar';
```
veej: It this working? you're not importing `NavBarProps`

ascariandrea: I think is better write as
```js
import NavBar, { NavBarProps } from './NavBar';
export default NavBar;
export { NavaBarProps };
```

but it's just a personal opinion (and for consistency with other components)

src/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }),
+  fixed: t.maybe(t.Boolean),
+  height: t.maybe(t.union([t.String, t.Number])),
+  background: t.maybe(t.String),
+  className: t.maybe(t.String),
+  style: t.maybe(t.Object)
+};
+
+@props(Props)
+export default class NavBar extends React.PureComponent<NavBarProps> {
```
ascariandrea: It wasn't a `PureComponent` before

src/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,84 @@
...
+import cx from '../utils/classnames';
+import { props, t, ReactChildren } from '../utils';
+import FlexView from 'react-flexview';
+
+export type NavBarProps = {
+  /** Left content: It doesn't shrink nor grow */
+  /** Center content. It grows to use all the available space */
+  /** Right content. It doesn't shrink nor grow */
+  /** Max Width. For css `max-width` */
+  content: {
```
ascariandrea: Are you sure about having all the comments before the object definition?

src/NavBar/NavBar.tsx
FrancescoCioria: @francescogior he's right, but we've already changed most of them during the other refactors so I would leave it like this (we're going to kill `@pure` anyway before a release)

src/NavBar/NavBar.tsx
francescogior: As far as I know, there is a typings called `ReactChild` that we could use it but we didn't agree yet and we are typing children everywhere with `any` (and add a `TODO`)

```ts
type ReactChild = ReactElement<any> | ReactText;
```

\cc @FrancescoCioria @giogonzo

src/NavBar/NavBar.tsx
francescogior: It's because I don't know if I can comment inner properties @giogonzo @FrancescoCioria

giogonzo: I think so, anyway you can easily test this "rendering" the component, even from the same file, and see what the VSCode completion popup/tooltip shows you

src/NavBar/NavBar.tsx
giogonzo: these are not children though, are other correctly typed props, amiright?

src/NavBar/NavBar.tsx
francescogior: tested, comments in inner props are not supported

src/NavBar/NavBar.tsx
francescogior: they were typed at aruntime as `tcomb.ReactChildren anyway

from @types/react/index.d.ts
```ts
    // React Nodes
    // http://facebook.github.io/react/docs/glossary.html
    // ----------------------------------------------------------------------

    type ReactText = string | number;
    type ReactChild = ReactElement<any> | ReactText;

    // Should be Array<ReactNode> but type aliases cannot be recursive
    type ReactFragment = {} | Array<ReactChild | any[] | boolean>;
    type ReactNode = ReactChild | ReactFragment | boolean | null | undefined;

    //
```

I'd take `ReactNode[]`

francescogior: But trying to assign a string to `content.left` would get me 
```
Type 'string' is not assignable to type 'ReactNode[] | undefined'.
```

🤔

veej: I think that's why you declared `children` as an *array* of ReactNodes... try to declare it as `ReactNode` an pass an array of children

FrancescoCioria: It should be `ReactNode`, not `ReactNode[]`. This is the definition of `ReactNode`:

```ts
type ReactText = string | number;
type ReactChild = ReactElement<any> | ReactText;

// Should be Array<ReactNode> but type aliases cannot be recursive
type ReactFragment = {} | Array<ReactChild | any[] | boolean>;
type ReactNode = ReactChild | ReactFragment | boolean | null | undefined;
```

`ReactNode` already has `any[]` through `ReactFragment`.

(BTW, `t.ReactChildren` accepts a single child too so even if it was a single node, the `[]` should be removed)

francescogior: > It should be ReactNode, not ReactNode[]. This is the definition of ReactNode:

yes my bad

`ReactNode` seems fine in the end, shouldn't be ok to use it in general for `children` or am I missing something?

FrancescoCioria: > ReactNode already has any[] through ReactFragment.

oops, actually it's a `Array<any[]>`

so, if you want to allow the users to pass `left: [<div />, 'poverone']` you should do `ReactNode | ReactNode[]`... but... are we sure we should support that weird case?

src/NavBar/NavBar.tsx
FrancescoCioria: ok, but this way we don't have a comment for `content`. I would add one

francescogior: No, the comment is the 4 lines paragraph and it's supported by vscode

PR #1114, opened 2017-09-01
#1066: [Typo] Dropdown (closes #1066)

### LINKED ISSUE #1066: [Typo] Dropdown
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4435899)

← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

### LINKED ISSUE #1127: Dropdown: drop support for `valueLink`
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4569607)

## requirements
`valueLink` is deprecated. we should remove it for every component, starting from `Dropdown`

## specs
remove `valueLink` prop

## misc
{optional: other useful info}

giogonzo:
Closes #1066 
Closes #1127 

@veej @FrancescoCioria In the end I decided to keep using the (quite broken) react-select types, and didn't remove simpleValue handling (so that this PR is almost non-breaking, removing valueLink only)
To be continued...

## Test Plan

- tests are happy
- TS is happy on Alinity PRO (couldn't test the app yet, due to other BRC-TS issues)

- [ ] Google Chrome
- [ ] Internet Explorer


francescogior:
ping @veej for `simpleValue`

veej:
`simpleValue` has default `true`, so we're using it everywhere we're not forcing it to `false` (and I'm not doing it on GDSM unless there is a real need, like in `SuggestionInput`).
We chose this default to be backward compatible, so I suppose it's highly breaking.

src/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,341 @@
...
+  };
+
+  valueToOption = (value: Props['value'], options: Select.Options) => {
+    if (t.String.is(value) || t.Number.is(value)) {
+      const { multi, delimiter } = this.props as DefaultedProps;
+      if (multi) {
+        const values = String(value).split(delimiter);
+        return values.map(v => find(options, { value: v })!); // TODO: `!`
+      }
+      return find(options, { value })!; // TODO: `!`
```
giogonzo: we were already implicitly assuming `value` as a valid option here, not sure how to fix this gracefully

veej:
@FrancescoCioria I added few small changes and fixed tests (see last three commits).
Changes to `valueToOption` are not relevant, since the function will be removed along with `simpleValue` prop. 
More important fixes are the `Value` definition and `onChange` type.

PR #1116, opened 2017-09-01
#1113: [Typo] Tooltip (closes #1113)

### LINKED ISSUE #1113: [Typo] Tooltip
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4476424)

← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

francescogior:
Closes #1113

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/Tooltip/TooltipTouch.tsx
```diff
@@ -22,6 +32,10 @@ export default class TooltipTouch extends React.PureComponent {
     isOpen: false
   };
 
+  private ref: HTMLDivElement
+
+  timeout: NodeJS.Timer
```
veej: this should be `: number`... please, make sure you're using `window.setTimeout` instead of `setTimeout` (by default, typescript is assuming we want to use the `setTimeout` from Node)

src/Tooltip/TooltipTouch.tsx
```diff
@@ -47,7 +61,7 @@ export default class TooltipTouch extends React.PureComponent {
     }, 200);
   }
 
-  onTouchMove = (e) => {
+  onTouchMove = (e: React.TouchEvent<HTMLDivElement>) => {
```
veej: minor, but you can annotate this as `onTouchMove: React.TouchEventHandler<HTMLDivElement>` and avoid the annotation on `e`

src/Tooltip/TooltipTouch.tsx
```diff
@@ -80,40 +94,42 @@ export default class TooltipTouch extends React.PureComponent {
...
   }
 
-  getLocals({ popover, ...props }) {
+  render() {
+    const { popover: _popover, children, className, id, style, type, size } = this.props as TooltipTouchDefaultedProps;
     const { isOpen } = this.state;
-
-    return {
-      ...props,
+    const popover = Object.assign({
```
veej: I don't understand the need of `Object.assign` here

src/Tooltip/TooltipTouch.tsx
```diff
@@ -80,40 +94,42 @@ export default class TooltipTouch extends React.PureComponent {
...
-        className: cx(popover.className, 'tooltip-touch')
-      },
-      spanStyle: isOpen ? {
-        display: 'hidden',
-        position: 'fixed',
-        top: -1000
-      } : undefined
+      className: cx(_popover.className, 'tooltip-touch')
+    });
+    const spanStyle: React.CSSProperties = {
```
veej: be aware: `spanStyle` was `undefined` if `!isOpen` before. Are you sure it's not introducing regressions?

src/Tooltip/TooltipTouch.tsx
```diff
@@ -80,40 +94,42 @@ export default class TooltipTouch extends React.PureComponent {
...
-        {...touchListeners}
+        onTouchStart={this.onTouchStart}
+        onTouchEnd={this.onTouchEnd}
+        onTouchMove={isOpen ? this.onTouchMove : undefined}
       >
         <span style={spanStyle}>{children}</span>
-        {isOpen && <Tooltip {...locals}>{children}</Tooltip>}
+        {isOpen && (
+          <Tooltip
+            id={id}
```
veej: mmm... I'm struggling to follow what's happening here after the removal of `@skinnable`, but it seems we were using a rest operator before and now we're not. Are we sure we're passing all the props? Isn't better to keep the "rest" logic we had before?

src/Tooltip/index.ts
```diff
@@ -0,0 +1,3 @@
+import Tooltip from './Tooltip';
+export default Tooltip;
+export { TooltipProps } from './Tooltip';
```
veej: very very minor, but just to stay consistent I would
```
import Tooltip, { TooltipProps } from './Tooltip';
export { TooltipProps };
```

src/Tooltip/TooltipTouch.tsx
gabro: almost: tsc correctly resolves window.sentiment, but vscode resolves the node one. Using window.setTimeout makes everyone happy

src/Tooltip/TooltipTouch.tsx
francescogior: it doesn't make sense at all ooops 💃

src/Tooltip/TooltipTouch.tsx
francescogior: I double checked and I found out that `className` wasn't there. Will remove. Nonetheless I think it's much more readable without spread operators and future refactors proof

veej: I don't know... Here `TooltipTouch` is using the same props defined for `Tooltip` (is not even defining a new ts type in this file), but doing this you're in some way listing the props `TooltipTouch` can receive.
What if you add a new optional prop to `Tooltip` and you completely forget about `TooltipTouch`? That prop will be ignored and you won't even notice it.

PR #1118, opened 2017-09-03
Make StatefulButton conform to #1082

gabro:
I must've missed this in the previous pass.

## Test plan
Tests still pass, it still works locally in the showroom.

### DIFF
--- a/src/Button/StatefulButton.tsx
+++ b/src/Button/StatefulButton.tsx
@@ -56,9 +56,9 @@ export interface StatefulButtonRequiredProps extends ButtonProps {
 
 export interface StatefulButtonDefaultProps {
   /** keep success state  */
-  stableSuccess?: boolean
+  stableSuccess: boolean
   /** time in millisecons to wait before state reset  */
-  timerMillis?: number
+  timerMillis: number
 }
 
 export type StatefulButtonProps = StatefulButtonRequiredProps & Partial<StatefulButtonDefaultProps>;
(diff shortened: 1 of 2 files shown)


test/components/StatefulButton.test.tsx
```diff
@@ -108,9 +108,7 @@ describe('StatefulButton', () => {
       const buttonInner = component.find('.button-inner');
       buttonInner.simulate('click');
       expect(buttonInner.hasClass('is-processing')).toBe(true);
-      const newProps = clone(component.props);
-      newProps.baseState = 'not-allowed';
-      component.setProps(newProps);
+      component.setProps({ baseState: 'not-allowed' });
```
gabro: not strictly related, but since I was touching `StatefulButton` I fixed this error.

src/Button/StatefulButton.tsx
```diff
@@ -137,15 +134,15 @@ export default class StatefulButton extends React.PureComponent<StatefulButtonPr
     this.timeoutId = setTimeout(() => {
       this.timeoutId = null;
       this.doResetInternalState();
-    }, this.getProps().timerMillis);
+    }, this.props.timerMillis);
```
veej: uhm 😕 
no need to cast to `DefaultedProps` here?

gabro: apparently not, I'll double check

gabro: it's not needed because `setTimeout` accepts `number | undefined`

PR #1119, opened 2017-09-03
No warnings in tests

gabro:
General cleanup of test output + I noticed Menu still had its tests in JS, so I converted them.

## Test plan
Watch the `npm test` output: no warnings.

image


test/components/StatefulButton.test.tsx
```diff
@@ -5,15 +5,33 @@ import StatefulButton, { StatefulButtonProps } from '../../src/Button/StatefulBu
...
 
 let consoleError: jest.SpyInstance<{}>;
+const consoleWarn = console.warn;
+
+const loadingSpinnerWarning =
+  `LoadingSpinner's parent node style should have "position: relative" or "position: absolute"`;
 
 beforeAll(() => {
   consoleError = jest.spyOn(console, 'error');
+  console.warn = jest.fn((...message) => {
```
gabro: an alternative to this would've been to set `style: 'relative'` in JSX in the `Button` implementation, but this seemed less breaking.

test/components/Meter.test.tsx
```diff
@@ -31,10 +31,9 @@ const exampleProps = {
   style,
   value: 25,
   min: 0,
-  max: 50,
+  max: 100,
```
gabro: this is the fix that required the least effort. Tweaking the ranges would've been more invasive.

src/DropdownMenu/Menu.tsx
```diff
@@ -94,7 +94,7 @@ export default class Menu extends React.PureComponent<MenuProps, MenuState> {
         <FlexView grow shrink className='menu-item-title'>
           {option.title}
         </FlexView>
-        <FlexView grow shrink={false} className='menu-item-metadata' hAlignContent='right'>
+        <FlexView shrink={false} className='menu-item-metadata' hAlignContent='right'>
```
gabro: @FrancescoCioria double check this makes sense. It's potentially breaking.

FrancescoCioria: I tested and it seems to work 👍

gabro:
@FrancescoCioria restored. I would also open a defect.

PR #1132, opened 2017-09-08
#1131: [Typo] TimePicker (closes #1131)

### LINKED ISSUE #1131: [Typo] TimePicker
← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

giogonzo:
Closes #1131

### based on https://github.com/buildo/react-components/pull/1114

This file is a complete mess. There are many TODOs around, I didn't have enough time to refactor it properly. I think it is safe to merge it as is for the moment

## Test Plan

- run tests

- [ ] Google Chrome
- [ ] Internet Explorer


src/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,343 @@
...
+  optionRenderer?: (option: Select.Option) => JSX.Element | null | false
+  /** called when the value of the `input` is changed */
+  onInputChange?: (inputValue: string) => void
+  /** called when dropdown is focused */
+  onFocus?: React.FocusEventHandler<HTMLDivElement>
+  /** called when dropdown is blurred */
+  onBlur?: React.FocusEventHandler<HTMLDivElement>
+  /** whether it should clear the input box on blur */
+  onBlurResetsInput?: boolean
+  /** wheter it should clear the input box on close */
```
gabro: Typo: wheter

test/components/Dropdown.test.js
```diff
@@ -37,49 +38,49 @@ describe('Dropdown', () => {
...
-      const { value } = dropdown.getLocals();
-      expect(typeof value).toBe('object');
-      expect(value).toEqual(exampleProps.value);
-    });
+  it('doesn\'t alter object value', () => {
+    const component = shallow(
+      <Dropdown {...exampleProps} />
+    );
+    const value = component.prop('value');
+    expect(typeof value).toBe('object');
```
gabro: This seems useless in TS

test/components/Dropdown.test.js
```diff
@@ -37,49 +38,49 @@ describe('Dropdown', () => {
...
+    const options = [
+      { value: 'test', label: 'Test' },
+      { value: 'test1', label: 'Test1' },
+      { value: 'test2', label: 'Test2' }
+    ];
+    const component = shallow(
+      <Dropdown {...Dropdown.defaultProps} value='test' options={options} />
+    );
+    const value = component.prop('value');
+    expect(typeof value).toBe('object');
```
gabro: aame

test/components/Dropdown.test.js
veej: it should not: `react-select` accepts numbers or strings as values, but we're forcing `Dropdown` to always pass an object. The test ensures that, when we pass a string or a number, it is correctly transformed into an object before passing it down to `react-select`

gabro: mm, ok, but is it the case in this test?

We pass `exampleProps`, whose `value` is an object.

test/components/Dropdown.test.js
gabro: @veej this makes more sense, we're testing a string is converted to an object.

test/components/Dropdown.test.js
veej: mm... well, the description says `it doesn't alter object value`, so I suppose the test is meant to ensure that the value is not transformed into a string or number.

gabro: ok, not a super interesting test then, but ok, I guess once we start using generics these problems are going away.

src/TimePicker/TimePicker.tsx
```diff
@@ -95,28 +102,28 @@ const maybeInsertSeparator = str => {
...
-  const hours = numberRegex.test(_hours) ? parseInt(_hours, 10) : '';
-  const minutes = !_minutes || numberRegex.test(_minutes) ? parseInt(_minutes, 10) : '';
+  const hours = numberRegex.test(_hours) ? parseInt(_hours, 10) : null;
+  const minutes = !_minutes || numberRegex.test(_minutes) ? parseInt(_minutes, 10) : null;
   const validHours = isValidHoursInTimeFormat(hours, timeFormat);
   const validMinutes = Minute.is(minutes);
 
   if (validHours && validMinutes) {
-    return { hours, minutes, originalInput: cleanedInput };
+    return { hours: hours!, minutes: minutes!, originalInput: cleanedInput };
```
gabro: are these safe?

src/TimePicker/TimePicker.tsx
```diff
@@ -131,41 +138,72 @@ export const createTimeList = ({ hours, minutes }, timeFormat) => {
...
+    originalInput: (time as any).originalInput || '', minTime, maxTime // TODO(typo)
   }));
   return uniqBy(sortBy(filteredTimeList.map(toOption), 'value'), 'value');
 };
 
+export type OnChangeInput = {
+  originalInput?: string
+} & Partial<Time>;
+
+export interface RequiredProps {
```
gabro: the naming is different than the conventions we're using other components, see #1082 

Also we've decided to export types, not interfaces.

src/TimePicker/TimePicker.tsx
veej: yep. They're checked in `isValidHoursInTimeFormat` and `Minute.is(minutes)`

gabro: mm, then we shouldn't need this if we use TS properly. I'll give it a look

PR #1135, opened 2017-09-15
#1134: [Typo] Tablo  (closes #1134)

### LINKED ISSUE #1134: [Typo] Tablo
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-4677211)

← #1034

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

veej:
Closes #1134

### tests performed
Since `JSX` types for children are quite broken, I tried to reach a stable version at least using children as functional components:

```
<Tablo data={data}
  {Column<DataType, 'foo'>({
    name: 'foo',
    children: [
      <Header>Foo</Header>,
      Cell<DataType, 'foo'>({ children: (data, rowData) => ... })
    ]
  }
</Tablo>
```

In this way a managed to have correct types even for `data` and `rowData` in `Cell` child function (with `data: undefined` in case of complex columns, where column name is not an attribute of `DataType`).

Notice that, by default, you can only use simple columns (where columns' names are `keyof T`), and `Tablo` will complain if you use unknown colums. In case you need to use compound columns, you can redefine `Tablo` like this:

```
type DataType = { foo: string };
type ColumnNames = keyof DataType | 'bar'
const MyTable: (props: TabloProps<DataType, ColumnNames>) => React.ReactElement<TabloProps<DataType, ColumnNames>> = Tablo;

<MyTable data={data}>
...
</MyTable>
```


gabro:
@veej can you rebase on master to get the new shiny CI?

src/Tablo/Cell/Cell.tsx
```diff
@@ -0,0 +1,95 @@
+import * as React from 'react';
+import * as cx from 'classnames';
+import { props, t, ReactChildren } from '../../utils';
+import FlexView from 'react-flexview';
+import { Cell as CellFDT } from 'fixed-data-table-2';
+import { If, StringContains, StringIntersection } from 'typelevel-ts';
+
+export type PickIfExists<T, K extends string> = If<StringContains<keyof T, K>, T[StringIntersection<keyof T, K>], undefined>;
+
```
francescogior: wouldn't be simply
```
export type PickIfExists<T, K extends string> = If<StringContains<keyof T, K>, T[K], undefined>;
```
?

francescogior: Also, I don't get what would happen in these cases
image

FrancescoCioria:
All these `T` and `K` are confusing... can't we rename them with a longer more meaningful name?

src/Tablo/Tablo.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /** callback to be called when the currently selected rows change */
+  onRowsSelect?: (selectedRows: number[]) => void,
+  /** single = only one selected row at a time, multi = multiple selection, none = selection disabled */
+  selectionType?: 'single' | 'multi' | 'none',
+  /** the id of the hovered row */
+  hoveredRowIndex?: number,
+  /** callback to be called when the hovered row changes */
+  onHoverRowChange?: (rowIndex: number) => void,
+  /** id of the column according which the data should be ordered */
+  sortBy?: keyof T,
```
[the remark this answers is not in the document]
FrancescoCioria: yes, I encountered the same error and using `K` instead of `keyof T` fixes it (it was missing the additional keys that can be defined in `K`).
I'll fix it in a commit

FrancescoCioria:
Ok, the build script was faulty and wasn't moving the `.d.ts` files from any subfolder to the relative subfolder in `lib` => in alinity pro we didn't have the typings for `Column`, `Cell` etc.

Once I fixed this (and the other issue pointed by @giogonzo) I finally managed to have a file with `Tablo` that typechecks and compiles 🎉 

(even if the code is veeery boilerplaty...)

src/Tablo/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+export { Column, ColumnProps };
+import Cell, { CellProps } from './Cell';
+export { Cell, CellProps };
+import Header, { HeaderProps } from './Header';
+export { Header, HeaderProps };
+import Footer, { FooterProps } from './Footer';
+export { Footer, FooterProps };
+
+const Component = autosize(columnsResize(columnsReorder(scrollable(selectable(sortable(_Tablo))))));
+export default function Tablo<T, K extends keyof T>(props: TabloProps<T, K>): React.ReactElement<TabloProps<T, K>> {
```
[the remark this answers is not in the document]
giogonzo: (if the idea is this is the default with only "standard" columns that are keys of `T`: probably `K extends string = keyof T`)

src/Tablo/Cell/Cell.tsx
giogonzo: > wouldn't be simply

Unfortunately not: in `If<a, b, c>`, `K` cannot yet be used in `b` position as `T[K]` since `K extends keyof T` is not true there. `StringIntersection` makes it true

giogonzo:
Just noticed that some of the remaining boilerplate can be removed pretty easily: once you have `MyTable` (= `Tablo<DataType, Colums>`), all the `children` type parameters needed for `Column` and `Cell` can indeed be inferred by TS:
- `T` is available because of the `children` typing (in both cases: `Tablo.children: Column[]` and `Column.children: Cell[]`)
- `K` can be easily inferred since it is an argument: `Column.name`

So I think @veej your very first example in this PR description can be simplified (didn't try but you shouldn't need to pass `<DataType, 'foo'>` to `Column` and `Cell`)

This is the resulting usage for the same AlinityPRO table: https://github.omnilab.our.buildo.io/buildo/alinity-pro/blob/dc75dec829231027c8c4a6110463e18c7587dfe1/qia/web/src/app/components/AlertCenterGrid/bsq/AlertCenterGrid.tsx#L64-L133 , which I think is acceptable.

The only thing I'm not quite getting is why the first argument of `Cell.children` is not inferred correctly, e.g.:
```tsx
{Column({
  name: 'category',
  width: columnsWidths.category || 90,
  children: [
    <Header key='0'><TextOverflow label={this.formatMessage('AlertCenterGrid.bsq.category')} /></Header>,
    Cell({
      hAlignContent: 'center',
//
//   can't directly use `_` as `category` here because it is inferred as `string | number | Date | undefined`
//   even weirder, IntelliSense gives me the correct `Critical | Alert | Notification` hovering on the first arg :/
//               v
      children: (_, { category }) => <CategoryCell category={category} />
    })
  ]
})}
```

veej:
>>> The only thing I'm not quite getting is why the first argument of Cell.children is not inferred correctly

That's the reason why I explicitly added generics to columns and cells. For some reason, inference is buggish on children, and adding explicit generics is the only way I found to be sure everything has the right type

veej:
@giogonzo, I fixed all remaining `K extends keyof T` in `K extends string` (defaulted to `keyof T` only in component/plugins declaration).
Are we ready for merging?

PR #1138, opened 2017-09-24
New drone

gabro:
Migrate to Drone 0.8 syntax.

I also had to fix a failing snapshot test in order to make sure this works.

Please note that until this PR is merged there will be two CI running on this repo (that's why you'll see two status checks)


FrancescoCioria:
image

I can see everything even if I'm not logged in, is it correct?

Also, the log is veeery short... isn't it missing a lot of steps?

PR #1141, opened 2017-10-09
#644: TimePicker should allow to "localize" am/pm (closes #644)

### LINKED ISSUE #644: TimePicker should allow to "localize" am/pm
## description

Currently `am/pm` strings are hardcoded in time picker code
and there's no way to customize them from outside
## how to reproduce
## specs

either use `Intl` internally or accept a "timeFormatter" or both
## misc

{optional: other useful info}

francescogior:
Closes #644

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "stylelint-webpack-plugin": "^0.7.0",
     "ts-jest": "20.0.10",
     "ts-loader": "^2.3.3",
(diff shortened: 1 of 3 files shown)


francescogior:
> So we had a timeFormatter prop already but it wasn't used?

no

PR #1145, opened 2017-10-13
#1083: Use TS namespaces for exported types (closes #1083)

### LINKED ISSUE #1083: Use TS namespaces for exported types
← #1034

## requirements
Currently we're "namespacing" manually by prefixing the component name to the exported types.

We want to use a proper TS namespace instead. 

## specs
Instead of

```tsx
export type FooProps = {};
export default Foo extends React.Component<FooProps>;
```

we make it

```tsx
export namespace Foo {
  export type Props = {};
};
export default Foo extends React.Component<Props>;
```

From the user perspective, instead of

```tsx
import Foo, { FooProps } from 'react-foo';

type Alias = FooProps;
```

will do

```tsx
import Foo from 'react-foo';

type Alias = Foo.Props;
```

## misc
This is a **breaking** change, since we're changing the name of all exported types.

FrancescoCioria:
Closes #1083

### tests performed

TS is happy 🎉 

- [ ] Google Chrome
- [ ] Internet Explorer


FrancescoCioria:
`tsc --noEmit` complains with these errors:

image

FrancescoCioria:
tested on alinity-pro (branch `typo-brc`): everything works fine apart from two conflicting `@types/react`

PR #1165, opened 2017-12-29
#1164: Popover does not work on React16 when attached to body (closes #1164)

### LINKED ISSUE #1164: Popover does not work on React16 when attached to body
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-6390178)

## description
Popover does not work on React16 when attached to body

## how to reproduce
locally
- npm start (it uses react16 locally)
- open Dropdown examples and dev console
- trigger the one with `attachToBody={true}` -> throws the error:

image

## specs
...

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1164

### tests performed
example with `attachToBody` works both on React16 and React15

- [ ] Google Chrome
- [ ] Internet Explorer

### DIFF
--- a/src/Popover/Popover.tsx
+++ b/src/Popover/Popover.tsx
@@ -113,7 +113,7 @@ export class Popover extends React.Component<Popover.Props, State> {
   private containerNode: Element | null;
   private onMouseEventDebouncedWhenOpen: ((_: string) => void) & _.Cancelable | null;
   private onMouseEventDebouncedWhenClosed: ((_: string) => void) & _.Cancelable | null;
-  private ContextWrapper: React.ComponentType<{ context?: { [_: string]: any }, children: any }>;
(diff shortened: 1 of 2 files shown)


src/Popover/Popover.tsx
```diff
@@ -315,13 +314,18 @@ export class Popover extends React.Component<Popover.Props, State> {
     const hiddenPopover = this.getHiddenPopover();
     const { context } = this.getPopoverProps();
     const { ContextWrapper } = this;
-    ReactDOM.render(<ContextWrapper context={context}>{hiddenPopover}</ContextWrapper>, this.containerNode);
 
-    // add pointer to popover node
-    this.popoverNode = this.containerNode.children[0];
+    // render Popover and save popover size in "onMount" cb (visible popover will be rendered in componentDidUpdate)
```
giogonzo: did you read this https://github.com/facebook/react/issues/10309#issuecomment-318434635 ?
Seems slightly simpler (but same thing at an higher level)

PR #1210, opened 2018-03-29
#1207: Add a basic Input component (closes #1207)

### LINKED ISSUE #1207: Add a basic Input component
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-8547126)

## requirements
{describe the new feature}

## specs

## misc
{optional: other useful info}

giogonzo:
Closes #1207

### tests performed

screen shot 2018-03-29 at 13 02 25
screen shot 2018-03-29 at 13 02 04

- [x] Google Chrome
- [ ] Internet Explorer


src/Input/Input.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { props, t, ReactChildren } from '../utils';
+import omit = require('lodash/omit');
+import InputChildren from 'react-input-children';
+import { ObjectOverwrite } from 'typelevel-ts';
+
+export type InputRequiredProps = ObjectOverwrite<React.HTMLProps<HTMLInputElement>, {
+  /** value */
+  value: string,
+  /** onChange */
+  onChange: (value: string) => void,
```
FrancescoCioria: is there a case where we want a static input with only `defaultValue`? (maybe in addition with `disabled` `true`)

if yes, maybe we should make `value` and `onChange` optional? what do you think?

src/Input/Input.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const {
+      _onChange: onChange
+    } = this;
+
+    const inputProps = {
+      ...omit(props, 'onChange'),
+      value,
+      onChange,
+      wrapper: {
+        className: cx('input', { 'empty-value': value === '' }, className)
```
FrancescoCioria: in other components (like `Dropdown`) we did the opposite: `has-value`

src/Input/input.scss
```diff
@@ -0,0 +1,50 @@
...
+$height: 28px !default;
+$padding-h: 12px !default;
+$font-size: 14px !default;
+$font-weight: $brc-regular !default;
+$placeholder-color: $brc-coolGrey !default;
+$color: $brc-darkGrey !default;
+$border-color: $brc-silver !default;
+$color-focus: $brc-darkGrey !default;
+$border-color-focus: $brc-waterBlue !default;
+$disabled-opacity: .6 !default;
```
FrancescoCioria: this should be named `opacity-disabled` :P

src/Input/Input.tsx
giogonzo: static input -> `value='string' & onChange=() => {}`
static input with "placeholder" style... 🤷‍♂️ 

I'd leave it as is for the moment

src/Input/Input.tsx
giogonzo: done in last commit

src/Input/input.scss
giogonzo: fixed!

src/Input/Input.tsx
```diff
@@ -48,18 +56,34 @@ export class Input extends React.PureComponent<Input.Props> {
...
       value,
+      status,
+      wrapper = {},
       ...props
     } = this.props as InputDefaultedProps;
     const {
       _onChange: onChange
     } = this;
 
+    const isSuccess = status === InputStatusT('success');
```
FrancescoCioria: why `InputStatusT('...')`? what happened to TS? :P

giogonzo: 😅 copy-pasta

PR #1213, opened 2018-03-29
#1212: Add PasswordInput component (closes #1212)

### LINKED ISSUE #1212: Add PasswordInput component
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-8552800)

## requirements
{describe the new feature}

## specs
it should be based on `Input`

## misc
{optional: other useful info}

giogonzo:
Closes #1212

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


giogonzo:
@FrancescoCioria fixed tests

src/PasswordInput/Examples.md
```diff
@@ -0,0 +1,15 @@
+### Examples
+
+```js
+initialState = { value: '' };
+
+const onChange = value => setState({ value });
+
+<form className='ui form'>
```
FrancescoCioria: this is a left-over from when we used semantic's css, you can remove the `className` and, if you want, the `form` as well

src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export namespace PasswordInput {
+  export type Props = PasswordInputRequiredProps & Partial<PasswordInputDefaultProps>;
+}
+type PasswordInputDefaultedProps = PasswordInputRequiredProps & PasswordInputDefaultProps;
+
+export const Props = {
+  ...InputPropsT,
+  hideText: t.maybe(t.String),
```
FrancescoCioria: can you add a description for these two new props?

src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+    this.setState(({ show }) => ({ show: !show }));
+  }
+
+  render() {
+    const props = this.props as PasswordInputDefaultedProps;
+    const type = this.state.show ? 'text' : 'password';
+    const text = this.state.show ? props.hideText : props.showText;
+    const toggleProps = {
+      onClick: this.onToggleClick,
+      className: 'toggle',
```
FrancescoCioria: this `className` is named as one of our components => `Toggle`'s CSS properties may be applied to this.
CSS modules may be a long-term solution for these problems, but for now I'd simply rename it

giogonzo:
@FrancescoCioria done!

PR #1238, opened 2018-05-04
#1237: Update all string refs (closes #1237)

### LINKED ISSUE #1237: Update all string refs
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-9472947)

## Description
Remove deprecated string refs and replace them with callback refs.

veej:
Closes #1237

### tests performed
Tests are still 🍏 
Components are still working

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -306,7 +306,7 @@ export class Dropdown extends React.Component<Dropdown.Props> {
     }
   }
 
-  private select: Select;
+  private select: Select | null;
   focus = () => { this.select && this.select.focus(); }
 
   optionGroupRenderer = (title?: string) => {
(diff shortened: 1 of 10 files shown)


giogonzo:
@veej did you investigate the `| undefined` on a callback ref?
The react doc say:
> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts. ref callbacks are invoked before componentDidMount or componentDidUpdate lifecycle hooks.

and: 
> If the ref callback is defined as an inline function, it will get called twice during updates, first with null and then again with the DOM element. This is because a new instance of the function is created with each render, so React needs to clear the old ref and set up the new one. You can avoid this by defining the ref callback as a bound method on the class, but note that it shouldn’t matter in most cases.

veej:
@giogonzo yep, I found the same thing, and I think it should not be a problem to consider the ref as non-null where we use it. What's your opinion?
A safer alternative could be to move the `!` wherever we use the ref (only if we're sure the component is mounted) of check for null every time we read it.

PR #1240, opened 2018-05-04
#1239: One `Menu` to rule them all (closes #1239)

### LINKED ISSUE #1239: One `Menu` to rule them all
## requirements
First of all, the name `DropdownMenu` is weird as it makes you think of `Dropdown` which is a completely different component.

Second, we should add the possibility to render the menu with the actions above instead of below

## specs
rename in `Menu`
add prop `position?: 'top' | 'bottom'` defaulted to `'bottom'`

FrancescoCioria:
Closes #1239

## Test Plan

- [ ] Google Chrome
- [ ] Internet Explorer


src/Menu/Menu.tsx
```diff
@@ -7,21 +7,21 @@ import { Popover } from '../Popover/Popover';
...
-  /** className for menu button icon (if children is passed, this is ignored) */
+  /** className for the buil-in icon rendered in the menu trigger (if children is passed, this is ignored) */
   iconClassName?: string,
-  /** called when menu is open */
+  /** called when menu is opened */
   onOpen: () => void,
   /** called when menu is closed */
   onClose: () => void,
-  /** the height of the menu button */
+  /** the a set of built-in max heights for the menu (if maxHeight is passed, this is ignored) */
```
giogonzo: `the a set`?

Also, generic comment: this kind of rules (key X is ignored if Y is passed) scream "union type"!

PR #1247, opened 2018-05-21
#1211: Tablo Column does not accept sortable prop (closes #1211)

### LINKED ISSUE #1211: Tablo Column does not accept sortable prop
## description
Passing `sortable` to `Column` results in a TS compilation error:

image

image

## how to reproduce

## specs

## misc
{optional: other useful info}

ecamellini:
Closes #1211

### tests performed


ecamellini:
@veej I'm not sure about this. What I see is that `sortable` works in the showroom examples, meaning that the fact that TS complains is probably only due to an error in the type interfaces that this PR should fix. Do you think this is sufficient?

ecamellini:
@veej I don't know if TS complains or not 😅I don't know how to test it. Should we maybe put it in both props (intrinsic and not)?

veej:
If `tsc` doesn't complain during the build of react-components itself, I think it's 👍

PR #1254, opened 2018-07-02
#1253: Deprecate Icon component in favor of SVG icons (closes #1253)

### LINKED ISSUE #1253: Deprecate Icon component in favor of SVG icons
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-10958347)

## requirements
Every component accepting an icon name should instead accept an SVG component as prop.

## specs

## misc
{optional: other useful info}

veej:
Closes #1253

### tests performed


FrancescoCioria:
@veej a general comment (I'll make a better review later on): I think you should use `t.ReactChild` (I'm not sure, just check) instead of `t.ReactChildren` in order to be consistent with TS typings and refuse `JSX.Element[]`

PR #1257, opened 2018-07-06
#1255: Implement FormField (closes #1255)

### LINKED ISSUE #1255: Implement FormField
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-11127193)

## requirements
Implement a wrapper for input components in a form, in order to handle labels, "required" flag, "disabled" flag, etc.

## specs
- The component should expose the following props:
  - `label` (note: use `htmlFor` in order to focus the component when the user clicks on the label)
  - `required` => put a `*` before the label if the field is required
  - `children`: the input component
  - every `FlexView` prop, that should be forwarded to the wrapper View
- Implement useful extensions for every common field (i.e. `InputField`, `DropdownField`, `CheckboxField`, ...) taking the props of the children component (the children should not be explicitly passed)

## misc
{optional: other useful info}

veej:
Closes #1255

### tests performed
image


src/FormField/DatePickerField.tsx
```diff
@@ -0,0 +1,53 @@
...
+    required?: boolean,
+    /** optional props to pass to the wrapping View */
+    viewProps?: View.Props,
+    /** an optional class name to pass to top level element of the component */
+    className?: string,
+    /** an optional style object to pass to top level element of the component */
+    style?: React.CSSProperties,
+    /** an optional id passed to the input component */
+    id?: string
+  } & ObjectOmit<DatePicker.Props<T>, 'returnFormat'>;  /* returnFormat is causing issues with the inference of OnChangeProps */
```
veej: Since TS 2.6.2 is not able to correctly infer the type for `onChangeProps` (union type for `returnFormat` and `onChange`), I dropped the support for one of the two APIs (it won't be possible to specify a `returnFormat` and receive the new value as string in `onChange`).
Newer versions of TS should be able to infer this correctly, but they break other things.

src/FormField/FormField.tsx
```diff
@@ -0,0 +1,62 @@
...
+  render() {
+    const { label, required, disabled, children, className: _className, fieldId, viewProps: _viewProps } = this.props;
+    const className = cx('form-field', _className, { 'is-disabled': disabled, 'is-required': required });
+    const viewProps = {
+      column: true,
+      ..._viewProps
+    };
+
+    return (
+      <View {...viewProps} grow className={className}>
```
giogonzo: `grow` is "hardcoded" here.. should we move it among other `viewProps` same as `column`?

veej:
@giogonzo I tried to add a `dropdownRenderer: (props: Dropdown.Props) => JSX.Element` prop, defaulted to `(props) => <Dropdown {...props} />` and could work fine.

PR #1276, opened 2018-10-04
#1275: Divider: accept custom classnames (closes #1275)

### LINKED ISSUE #1275: Divider: accept custom classnames
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-13564033)

## requirements
Currently `Divider` doesn't accept any custom classNames, making it difficult to customize it.

## specs

## misc
{optional: other useful info}

armoniaechaos:
Closes #1275

### tests performed

### DIFF
--- a/src/Divider/Divider.tsx
+++ b/src/Divider/Divider.tsx
@@ -10,13 +10,12 @@ export type DividerDefaultProps = {
   /** an optional style object to pass to top level element of the component */
   style: React.CSSProperties,
   /** an optional class name to pass to first inner element of the component */
-  className?: Divider.className;
+  className?: string;
 };
 
 export namespace Divider {
   export type Orientation = 'horizontal' | 'vertical'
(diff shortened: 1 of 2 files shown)


src/Divider/Divider.tsx
```diff
@@ -9,12 +9,14 @@ export type DividerDefaultProps = {
...
   style: React.CSSProperties,
+  /** an optional class name to pass to first inner element of the component */
+  className?: Divider.className;
 };
 
 export namespace Divider {
   export type Orientation = 'horizontal' | 'vertical'
   export type Size = 'small' | 'medium' | 'large' | 'no-margin'
-
+  export type className = string;
```
FrancescoCioria: we're creating a type alias for the the other two because they are complex types, but for a simple string is not necessary.

please remove this and replace `Divider.className`  above with `string`

PR #1277, opened 2018-10-05
#1167: Drop support for React 15 (closes #1167)

### LINKED ISSUE #1167: Drop support for React 15
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-6584860)

## why?
By dropping support for React15 we'll be able to use some new cool features from React16 like:
- fragments
- portals
- error boundaries

## requirements
ensure every project at buildo is ready to upgrade to React16

## specs
change 1/2 lines of code in the `package.json` :)

## misc
{optional: other useful info}

FrancescoCioria:
Closes #1167

### tests performed


src/Toaster/Toaster.tsx
```diff
@@ -72,7 +72,7 @@ export class Toaster extends React.Component<Toaster.Props> {
   componentDidMount() {
     const node = this.props.attachTo
       ? this.toaster!
-      : ReactDOM.findDOMNode(this).parentElement!;
+      : (ReactDOM.findDOMNode(this) as Element).parentElement!;
```
giogonzo: reuse the utility here?

test/components/Tablo.typecheck.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from "react";
```
giogonzo: is this file intended to make the build fails if Tablo doesn't typecheck?

FrancescoCioria: ups, I was using this to test `Tablo`'s types in development, it wasn't supposed to be committed

PR #1286, opened 2018-12-24
#781: Implement RadioGroup (closes #781)

### LINKED ISSUE #781: Implement RadioGroup
## requirements
This component should be able to render radio buttons in a completely custom layout. Each radio button should be correctly on/off based on the state passed to its parent (radiogroup).

See also https://zpl.io/a3qYxEy

## specs
{optional: other useful info}

## misc
{optional: other useful info}

giogonzo:
Closes #781

as a form field:
screenshot 2018-12-28 at 10 44 38

vertical:
screenshot 2018-12-28 at 10 44 24

horizontal:
screenshot 2018-12-28 at 10 44 32

### tests performed


src/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,99 @@
...
+  value: string;
+};
+
+export type RadioGroupRequiredProps = {
+  /** name for the group */
+  name: string;
+  /** value */
+  value?: string;
+  /** onChange */
+  onChange: (value: string) => void;
```
FrancescoCioria: is it the same as `Dropdown`? are we using the string value instead of the option?

FrancescoCioria:
@giogonzo I think we'd also like to use them stacked vertically instead of horizontally, is this possible with your code?

src/RadioGroup/RadioGroup.tsx
giogonzo: `Dropdown` is more complex: `export type Value = SelectNS.Option | SelectNS.Options | SimpleValue;`

I kept things as simple as possible here (only "simple value", and harcoded to `string`). Do you think we should add more flexibility?

giogonzo:
> I think we'd also like to use them stacked vertically instead of horizontally, is this possible with your code?

@FrancescoCioria I'll give it a try, should we default to horizontal or vertical?

src/RadioGroup/RadioGroup.tsx
FrancescoCioria: no, it's enough for now (also this is controversial in `Dropdown` as well...)

FrancescoCioria:
> @FrancescoCioria I'll give it a try, should we default to horizontal or vertical?

discussed with alessandra, the default should be vertical

src/RadioGroup/radioGroup.scss
```diff
@@ -69,4 +68,12 @@ $opacity-disabled: 0.6 !default;
       cursor: auto;
     }
   }
+
+  &:not(.is-horizontal) .radio-group-option:not(:last-child) {
+    margin-bottom: 8px;
+  }
+
+  &.is-horizontal .radio-group-option {
```
FrancescoCioria: missing `:not(:last-child)`?

PR #1303, opened 2019-01-18
#1285: Upgrade react-select to 2.x (closes #1285)

### LINKED ISSUE #1285: Upgrade react-select to 2.x
[Project card](https://github.com/buildo/react-components/projects/1?fullscreen=true#card-15867957)

fes300:
Closes #1285

## Test Plan


src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,111 @@
...
-  /** whether the menu should be rendered on top or bottom when it's open */
-  menuPosition: Dropdown.MenuPosition;
+  export type Props<OptionType> = Pick<
+    SelectNS.Props<OptionType>,
+    Exclude<keyof SelectNS.Props, "menuPosition">
+  > & { menuPosition: MenuPosition; size: "medium" | "small"; flat: boolean };
 }
 
 export const Props = {
-  value: t.maybe(t.union([t.Number, t.String, t.Object, t.list(t.Object)])),
```
veej: why value has been removed?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,111 @@
...
       "is-medium": size === "medium",
       "is-small": size === "small",
       "is-flat": flat,
-      "is-clearable": clearable,
+      "is-clearable": isClearable,
       "menu-position-top": menuPosition === "top"
     });
   }
 
-  _onChange = (_value?: Dropdown.Value | null) => {
```
veej: have you verified it is useless now?

src/Dropdown/Dropdown.tsx
fes300: error :)

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,111 @@
...
-      menuRenderer: this.menuRenderer as SelectNS.MenuRendererHandler
-    };
-
     return (
       <Select
-        {...selectProps}
+        {...props}
+        isClearable={isClearable}
+        backspaceRemovesValue={
+          t.Nil.is(backspaceRemovesValue) ? isClearable : backspaceRemovesValue
```
veej: I won't use `t.Nil.is` here. Moreover, if this is the prop coming from react-select, it seems it cannot be null.

src/Dropdown/Dropdown.tsx
fes300: I still have to check if this and `onInputKeyDown` are still relevant, in general I still have a lot of TODOs on this PR, just wanted to check if you like the approach

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,172 @@
...
       "menu-position-top": menuPosition === "top"
     });
   }
 
-  _onChange = (_value?: Dropdown.Value | null) => {
-    const value = _value === "" || isEmptyArray(_value) ? null : _value;
-    return this.props.onChange(value);
-  };
-
-  onInputKeyDown: SelectNS.OnInputKeyDownHandler = e => {
```
fes300: this was used to manage up and down keys with empty options. I don't know why it was originally implemented but seems useless now..

src/Dropdown/Dropdown.tsx
fes300: after some trials, I think this was due to `simpleValue` that is not supported anymore

veej:
There are also issue with the style (e.g. `isDisabled` doesn't change the style at all)

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,176 @@
...
     return (
       <Select
-        {...selectProps}
+        {...props}
+        isClearable={isClearable}
+        backspaceRemovesValue={
+          t.Nil.is(backspaceRemovesValue) ? isClearable : backspaceRemovesValue
+        }
+        components={{
+          Menu: props => (
```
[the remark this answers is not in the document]
veej: it's still possible to use `classNames` for the inner components passing a `classNamePrefix` prop to `react-select`: https://react-select.com/styles#using-classnames

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,176 @@
...
+                  "is-selected": props.isSelected
+                })
+              )}
+            />
+          ),
+          DropdownIndicator: props => (
+            <span className="Select-arrow-zone">
+              <span
+                className={
+                  props.selectProps.menuIsOpen
```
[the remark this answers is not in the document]
fes300: ops :)

src/Dropdown/Examples.md
```diff
@@ -1,184 +1,196 @@
...
   value={state.value}
   onChange={onChange}
-  searchable
-  clearable
-  backspaceRemoves
-  placeholder='Select some fruit (try to type "Banana")'
+  isSearchable
+  isClearable
+  backspaceRemovesValue
+  placeholder="Select some fruit (try to type &quot;Banana&quot;)"
```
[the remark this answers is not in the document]
FrancescoCioria: why not simply `\"`?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,176 @@
...
+            </components.MultiValueRemove>
+          ),
+          ClearIndicator: props => (
+            <span onClick={props.clearValue} className="Clear-indicator">
+              ×
+            </span>
+          ),
+          ...customComponents
+        }}
+        className={cx("buildo-dropdown", className, this.getCustomClassNames())}
```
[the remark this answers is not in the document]
FrancescoCioria: it's a good deterrent to not use a CSS framework like bootstrap :P

src/Dropdown/Examples.md
```diff
@@ -1,184 +1,196 @@
...
   value={state.value}
   onChange={onChange}
-  searchable
-  clearable
-  backspaceRemoves
-  placeholder='Select some fruit (try to type "Banana")'
+  isSearchable
+  isClearable
+  backspaceRemovesValue
+  placeholder="Select some fruit (try to type 'Banana';)"
```
FrancescoCioria: there's a `;` leftover

src/TimePicker/TimePicker.tsx
```diff
@@ -41,6 +43,21 @@ const isValidHoursInTimeFormat = (
...
 ): hours is number => (timeFormat === H24 ? Hour.is(hours) : Hour12.is(hours));
 
+const getComponents = (
+  timeFormatter?: TimePicker.TimeFormatter
+): SelectNS.Props<TimeDropdownOption>["components"] => {
+  return timeFormatter
+    ? {
+        Option: props => (
+          <components.Option {...props}>
+            {timeFormatter(props.data.time) || null}
```
FrancescoCioria: why `|| null` if `timeFormatter` returns a `JSX.Element`?

src/TimePicker/TimePicker.tsx
```diff
@@ -41,6 +43,21 @@ const isValidHoursInTimeFormat = (
...
+          </components.Option>
+        ),
+        SingleValue: props => timeFormatter(props.data.time) || null
```
FrancescoCioria: is it ok to pass ` JSX.Element` as `SingleValue`?
also, same comment as above: is `|| null` necessary?

src/TimePicker/TimePicker.tsx
```diff
@@ -310,25 +332,21 @@ export class TimePicker extends React.Component<
...
-      : undefined; // TODO(typo)
+    const components = getComponents(timeFormatter);
 
     return (
       <Dropdown
         {...{ id, className, style }}
-        searchable={searchable}
-        value={value}
+        isSearchable={searchable}
+        value={options.find(o => o.value === value)}
```
FrancescoCioria: 🚨 `.find` does NOT exist in IE11 🚨 
Bug!

please use `find` from `lodash`

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
-
-  export type MenuPosition = "top" | "bottom";
-
-  export type Props = RequiredProps & Partial<DefaultProps>;
-}
-
-type DefaultedProps = RequiredProps & DefaultProps;
+const defaultComponents: Dropdown.Props<any>["components"] = {
+  Placeholder: ({ children }) => (
+    <div className="react-select__placeholder">{children}</div>
```
FrancescoCioria: if we are re-defining it, why not using our own new class instead of `react-select__placeholder`?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
+      {...props}
+      className={cx({
+        "is-disabled": props.isDisabled,
+        "is-focused": props.isFocused,
+        "is-selected": props.isSelected
+      })}
+    />
+  ),
+  DropdownIndicator: props => (
+    <span className="react-select__select-arrow-zone">
```
FrancescoCioria: same: if we have to re-define them then I would implement a simple `Arrow` component instead of using their `className` and relying on their layout

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
-    props: MenuRendererProps
-  ) => JSX.Element[] | SelectNS.HandlerRendererResult;
-
-  export type MenuPosition = "top" | "bottom";
-
-  export type Props = RequiredProps & Partial<DefaultProps>;
-}
-
-type DefaultedProps = RequiredProps & DefaultProps;
+const defaultComponents: Dropdown.Props<any>["components"] = {
```
FrancescoCioria: It's not clear why we are overwriting some of them (like `Placeholder`).
could you add some comments in the code?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
+      innerProps={{
+        ...props.innerProps,
+        className: "react-select__multi-value-remove"
+      }}
+    >
+      ×
+    </components.MultiValueRemove>
+  ),
+  ClearIndicator: props => (
+    <span onClick={props.clearValue} className="react-select__clear-indicator">
```
FrancescoCioria: same, as we are defining our own component, I would use our own `className`

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
-      onInputKeyDown,
-      onChange: _onChange,
-      menuRenderer: this.menuRenderer as SelectNS.MenuRendererHandler
-    };
-
     return (
       <Select
-        {...selectProps}
+        {...props}
+        isClearable={isClearable}
```
FrancescoCioria: why is this explicit?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
+        backspaceRemovesValue={backspaceRemovesValue}
```
FrancescoCioria: why is this explicit?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,160 @@
...
   render() {
     const {
-      _onChange,
-      onInputKeyDown,
-      props: { className, options, backspaceRemoves, clearable, ...props }
+      props: {
+        className,
+        backspaceRemovesValue,
+        isClearable,
+        menuPosition,
```
FrancescoCioria: `menuPosition` is not used => never passed to react-select through `...props`

src/Dropdown/Dropdown.tsx
fes300: this class is indeed not defined by `react-select` (unfortunately some elements [like placeholder] are not given a class to target them anymore).
I used a name that would follow `react-select` convention as IMHO if we use `react-select` classes it would be messy to mix in other naming conventions

src/Dropdown/Dropdown.tsx
fes300: leftover :)
it was used to calculate `backspaceRemovesValue`

src/Dropdown/Dropdown.tsx
fes300: as before 👍

src/Dropdown/Dropdown.tsx
fes300: `menuPosition` is defined here with the explicit intention of not passing it to `react-select` as it is now an accepted prop but we want to keep our implementation (or maybe not? what do you think?)

src/TimePicker/TimePicker.tsx
fes300: this is a very good question for which I have no reasonable answer...

src/TimePicker/TimePicker.tsx
fes300: the first answer is "yes", the second is "not and I have no clue why I put this `||`"...

src/Dropdown/Dropdown.tsx
FrancescoCioria: ok, now it's clear 👍 

> (or maybe not? what do you think?)
what are the benefits of our own implementation compared to react-select one?

src/Dropdown/Dropdown.tsx
veej: > this class is indeed not defined by `react-select` (unfortunately some elements [like placeholder] are not given a class to target them anymore).

This is not true. Those components have custom classes (https://react-select.com/styles)

src/TimePicker/TimePicker.tsx
```diff
@@ -297,7 +297,7 @@ export class TimePicker extends React.Component<
 
   state = { inputValue: "" };
 
-  _onChange = (value?: TimeDropdownOption) => {
+  _onChange = (value: TimeDropdownOption) => {
     if (value) {
```
FrancescoCioria: is this `if` still necessary?

src/Dropdown/Dropdown.tsx
```diff
@@ -1,400 +1,110 @@
...
-        e.preventDefault();
-      }
+  isMulti = (): boolean => {
+    switch (this.props.type) {
+      case "multi":
+      case "multi-clearable":
+        return true;
+      case "single":
+      case "single-clearable":
+        return false;
```
giogonzo: with the `noFallthroughCasesInSwitch` option we are using in the default tsconfig in scriptoni, this would fail

src/Dropdown/Examples.md
```diff
@@ -1,184 +1,200 @@
 ### Examples
 
 ```js
-initialState = {}
```
giogonzo: looks like the majority of changes to this file are just due to prettier (so out of scope for the PR). There's always the option to "File: save without formatting" in vscode. Or, to prevent this happening again, we could setup a prettierrc for this folder I guess

PR #1312, opened 2019-02-22
Adding preview to CI

ecamellini:
Pre-requisites:
- [x] merging https://github.com/buildo/terraform/pull/236
- [x] applying terraform changes
- [ ] deploying new `github-workflow-pal` from `master` (not really required for the merge --> just check http://react-components.preview.our.buildo.io/adding-preview.html)
- [x] testing preview on this PR (pipeline already applied)

### DIFF
--- a/ci/tasks/build-preview.sh
+++ b/ci/tasks/build-preview.sh
@@ -7,7 +7,7 @@ cd react-components
 
 yarn install --no-progress
 
-yarn build-styleguidist
+yarn build-styleguidist 
 
 mv styleguide/index.html styleguide/$BRANCH.html
 
--- a/ci/tasks/build-preview.sh
+++ b/ci/tasks/build-preview.sh
@@ -13,8 +13,7 @@ cd react-components
 
 yarn install --no-progress
 
-yarn build-styleguidist
+NODE_ENV=production yarn build-styleguidist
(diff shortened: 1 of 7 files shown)


ci/tasks/build-preview.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+set -x
+
+cd react-components
+
+yarn install --no-progress
+
+ yarn build-styleguidist
```
gabro: extra space?

ecamellini:
It doesn't work, but it should 🤷‍♂️

ecamellini:
Found the problem, let's see

ecamellini:
http://react-components.preview.our.buildo.io/adding-preview.html it works! But...since this build does not produce a bundle with a different hash every time => all files are overwritten. This means that only the latest build will be shown in preview, and can be misleading.

Maybe it's not a problem? There are not many PRs in parallel on BRC.

See:
http://react-components.preview.our.buildo.io/adding-preview.html

ecamellini:
@gabro now it works: I managed to produce a bundle and I'm uploading to the bucket only index file and bundle files:

image

content of the build folder:

image

PR #1345, opened 2019-05-02
#1344: Form with "submit on enter" behavior (closes #1344)

### LINKED ISSUE #1344: Form with "submit on enter" behavior
Add a basic Form component with the only feature of "submitting on Enter press"

giogonzo:
Closes #1344

### tests performed

### DIFF
--- a/src/Form/Examples.md
+++ b/src/Form/Examples.md
@@ -1,3 +1,38 @@
 ### Examples
 
-This component is a work in progress, we suggest you to ignore it for now.
+Simple wrapper for form that adds "submit on Enter" behavior
+
+```js
+initialState = {
+  firstName: "",
+  lastName: ""
+};
+
+const onChange = key => value => setState({ [key]: value });
+
+<Form.Form
+  column
+  submitLabel="submit"
+  renderSubmit={props => (
+    <StatefulButton
+      {...props}
+      style={{ ...props.style, width: "300px", marginLeft: "auto" }}
+    />
+  )}
+  onSubmit={() => {
+    alert("submitted!");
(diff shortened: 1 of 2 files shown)


src/Form/Form.tsx
```diff
@@ -1,20 +1,74 @@
...
 }
 
-export const Props = t.Object;
-
-@props(Props)
 export class Form extends React.PureComponent<Form.Props> {
+  formRef = React.createRef<HTMLFormElement>();
+
+  onFormSubmit: React.ReactEventHandler<HTMLFormElement> = e => {
+    if (this.formRef.current) {
```
giogonzo: this code is temporary: `Form` should actually enforce the error guidelines in the future, such as avoid stateful buttons; it will thus optionally show an error and track the Promise state internally

src/Form/Examples.md
```diff
@@ -1,3 +1,38 @@
...
+const onChange = key => value => setState({ [key]: value });
+
+<Form.Form
+  column
+  submitLabel="submit"
+  renderSubmit={props => (
+    <StatefulButton
+      {...props}
+      style={{ ...props.style, width: "300px", marginLeft: "auto" }}
+    />
```
gabro: what if we have a form with a submit and a cancel button, for example a modal. This feels a bit limited for those cases, doesn't it?

giogonzo: Agreed, but you can always render
```ts
renderSubmit={submitProps => (
  <View><Cancel /><Submit {...submitProps} /></View>
)}
```

gabro: fair enough.

PR #1333, opened 2019-03-08
#1321: FormField Hint (closes #1321)

veej:
Closes #1321

### tests performed

**label**
image

**box**
image

**tooltip**
image

PR #1348, opened 2019-05-03
#1347: Remove returnFormat prop from DatePicker (closes #1347)

### LINKED ISSUE #1347: Remove returnFormat prop from DatePicker
@gabro is proposing to remove `returnFormat` (and `T`s other than `Date`) completely

_Originally posted by @giogonzo in https://github.com/buildo/react-components/pull/1333_

veej:
Closes #1347

### tests performed

Example in showroom still works as expected

### DIFF
--- a/src/DatePicker/DatePicker.tsx
+++ b/src/DatePicker/DatePicker.tsx
@@ -6,42 +6,27 @@ import { props, t, ReactChild } from "../utils";
 import * as moment from "moment";
 
 export namespace DatePicker {
-  export type Value = string | Date | moment.Moment;
-  export type OnChangeProps<T extends Value> =
-    | {
-        /** MomentJS format used to format date before returing through "onChange" */
-        returnFormat?: never;
-        /** called when value changes */
-        onChange?: (date?: T) => void;
-      }
-    | {
-        /** MomentJS format used to format date before returing through "onChange" */
-        returnFormat: string;
-        /** called when value changes */
-        onChange?: (date?: string) => void;
-      };
-
-  export type Props<T extends Value> = {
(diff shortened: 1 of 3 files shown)


giogonzo:
my bad, the comment in https://github.com/buildo/react-components/pull/1333#discussion_r280793367 wasn't really clear. I think @gabro already has a PR in the making including this change?

gabro:
@giogonzo not yet, I'm still working on the DateRangePicker component but I haven't touched the DatePicker yet.

src/DatePicker/DatePicker.tsx
```diff
@@ -6,50 +6,35 @@ import { props, t, ReactChild } from "../utils";
...
+    onChange?: (date?: Date) => void;
     /** called when datepicker is closed */
     onHide?: () => void;
     /** MomentJS format used to display current date */
     displayFormat?: string;
-    /** called when value is cleared */
-    onClear?: () => void;
     /** minimum date selectable by the user */
-    minDate?: Value;
+    minDate?: Date | string;
```
gabro: is there a reason to also keep `string` in input? Can't we just stick to `Date`

veej: no particular reason. I left it just for convenience: there's no complex logic around these inputs and you have to convert it to a `Moment` date anyway, so I decided to keep everything that can be passed to `moment()`.
btw I'm ok to remove them if you prefer.

gabro: I prefer removing it. Having a stricter interface will make our projects more consistent and it allows changing the internals more freely.

PR #1353, opened 2019-05-28
#1352: expose components prop in TimePicker (closes #1352)

### LINKED ISSUE #1352: expose components prop in TimePicker
TimePicker should allow the user to define a set of custom components to allow different styling.

fes300:
Closes #1352

### tests performed


src/TimePicker/TimePicker.tsx
```diff
@@ -259,6 +259,8 @@ export interface DefaultProps {
   size: "medium" | "small";
   /** whether the menu should open on top or bottom */
   menuPosition: Dropdown.Props<any>["menuPlacement"];
+  /** object of custom compoents for react select */
+  components: Dropdown.Props<MakeOptionsInput>["components"]
```
giogonzo: @fes300 we also need to update the runtime `Props` typing

PR #1358, opened 2019-06-03
Adding form errors

ecamellini:


ecamellini:
@giogonzo the first two commits I did could actually be considered "unrelated" from the form error management, but I would leave them as part of this work. Let's see what we get eventually.

PR #1368, opened 2019-09-13
#947: Dropdown: SelectAll in multi mode (closes #947)

### LINKED ISSUE #947: Dropdown: SelectAll in multi mode
## requirements
- When dropdown is in `multi` mode, it should be possible to add an "all" option that allows to select all the options at once.
- The "All" option can't be selected along with other options: when "all" is selected, it should be the only selected value (this means that all the previously selected options must be cleared), and the dropdown should have the style of a single-value selection (i.e. `All` shouldn't be shown as a tag)
- When `all` is selected, it should be possible to select any other value. When this happens, the new value should replace "all"

## misc
{optional: other useful info}

veej:
Closes #947

### DIFF
--- a/src/Dropdown/Examples.md
+++ b/src/Dropdown/Examples.md
@@ -69,7 +69,7 @@ const options2 = [
 When needed (e.g. when there isn't enough space under the dropdown to render the menu), it's possible to force the menu to open on top
 
 ```js
-intialState = {};
+initialState = {};
 
 const onChange = value => setState({ value });
 
@@ -95,7 +95,7 @@ const options = [
 - When the `isSearchable` prop is used, it's possible to search the desired value by writing into the dropdown.\n\* Use `isClearable` if it should be possible to reset the selected value clicking on the `delete` icon that will appear once a value is selected and `backspaceRemovesValue` if you want to be able to remove the last value selected with the backspace button.
 
 ```js
-intialState = {};
+initialState = {};
 
 const onChange = value => setState({ value });
 
@@ -129,7 +129,7 @@ const options = [
 Dropdown could also allow the selection of multiple values
 
 ```js
-intialState = {};
+initialState = {};
 
 const onChange = value => setState({ value });
 
@@ -156,9 +156,9 @@ const options = [
 Using a `MultiDropdownWithSelectAll` component, it's possible to automatically add an "all" option that will select all the available options at once.
 
 ```js
-intialState = {};
-const onChange = (value, isAll) => {
-  if (isAll) {
+initialState = {};
+const onChange = value => {
+  if (value.value === "_ALL") {
     console.log("ALL selected!");
   }
   setState({ value });
@@ -176,7 +176,7 @@ const options = [
   onChange={onChange}
   placeholder="Select some fruit(s)"
   options={options}
-  selectAll={{ label: "All" }}
+  selectAllLabel="All"
 />;
 ```
 
--- a/src/Dropdown/Examples.md
+++ b/src/Dropdown/Examples.md
@@ -151,6 +151,35 @@ const options = [
 />;
 ```
 
+#### Multiselect with "select all" option
+
+Using a `MultiDropdownWithSelectAll` component, it's possible to automatically add an "all" option that will select all the available options at once.
+
+```js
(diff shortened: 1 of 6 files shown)


src/Dropdown/Examples.md
```diff
@@ -151,6 +151,35 @@ const options = [
 />;
 ```
 
+#### Multiselect with "select all" option
+
+Using a `MultiDropdownWithSelectAll` component, it's possible to automatically add an "all" option that will select all the available options at once.
+
+```js
+intialState = {};
+const onChange = (value, isAll) => {
```
veej: I don't really like this API to check if `all` is selected. I need to play a little bit more with it

src/Dropdown/MultiDropdownWithSelectAll.tsx
```diff
@@ -0,0 +1,162 @@
...
+  getCommonClassnames,
+  DefaultProps
+} from "./commons";
+import { GroupType, ValueType } from "react-select/lib/types";
+
+export function allSelected<OptionType = never>(): SelectAllValue<OptionType> {
+  return { type: "AllSelected" };
+}
+
+export function someSelected<OptionType>(values: OptionType[]) {
```
giogonzo: just for coherence, i would add the return type annotation: `SelectAllValue<OptionType>`

src/Dropdown/MultiDropdownWithSelectAll.tsx
```diff
@@ -0,0 +1,162 @@
...
+  return { type: "SomeSelecetd", values };
+}
+
+type SelectAllOptionType = { value: "_ALL"; label: string };
+type SelectAllValue<OptionType> =
```
giogonzo: this should be exported as well since we are exporting the constructors

src/Dropdown/MultiDropdownWithSelectAll.tsx
```diff
@@ -0,0 +1,162 @@
...
+  | {
+      type: "AllSelected";
+    }
+  | {
+      type: "SomeSelected";
+      values: OptionType[];
+    };
+
+type NonDefaultProps<OptionType> = CommonProps<
```
giogonzo: does `CommonProps` really need the `OptionType` type param? Meaning, shouldn't all value/option related props be handled specifically by each variant outside of `CommonProps`? If it's only used internally to pass it to react select Props type, maybe we can use `unknown` there

src/Dropdown/MultiDropdownWithSelectAll.tsx
```diff
@@ -0,0 +1,162 @@
...
+    value: "_ALL",
+    label: this.props.selectAllLabel
+  };
+
+  injectSelectAllOptions = (
+    options: MultiDropdownWithSelectAll.Props<OptionType>["options"]
+  ):
+    | Array<OptionType | SelectAllOptionType>
+    | Array<GroupType<OptionType | SelectAllOptionType>> => {
+    if (options === undefined || options.length === 0) return [];
```
giogonzo: the `options === undefined` check should probably be removed?

src/Dropdown/MultiDropdownWithSelectAll.tsx
veej: There are many other props, like `components` or `options` that are not handled by specific cases. I tried to remove the type param, but I have many errors when I pass these props to react-select

src/Dropdown/MultiDropdownWithSelectAll.tsx
veej: `options` is defined as optional in react-select.
I'll overwrite its type in `CommonProps` to be `NonNullable`.

src/Dropdown/MultiDropdownWithSelectAll.tsx
giogonzo: if `options` is part of `CommonProps` then it should be `options: OptionType[]` i.e. not include `SelectAllOptionType` externally. Right?

veej: Yep, I noticed this few minutes ago and fixed it ;)

PR #1374, opened 2019-09-16
#1373: avoid using React.ReactNode in prop typings (closes #1373)

### LINKED ISSUE #1373: avoid using React.ReactNode in prop typings
as explained in https://github.com/buildo/react-flexview/pull/77
using a custom `Children` type (as defined there already) improves safety in UI rendering code

giogonzo:
Closes #1373

Updates `react-flexview` + TS dependencies (not strictly related but while at it..)

Introduces the `Children` type util and replaces usages of `React.ReactNode` with it

### tests performed

Tested on an internal project, didn't cause any issue


src/Popover/Popover.tsx
```diff
@@ -689,7 +689,7 @@ export class Popover extends React.Component<Popover.Props, State> {
     };
   };
 
-  getLocals() {
+  private getLocals() {
```
giogonzo: needed to suppress a weird type error

src/TextOverflow/TextOverflow.tsx
```diff
@@ -33,7 +33,7 @@ export type TextOverflowRequiredProps = {
   /** in case you want to use a custom component (like a `Tooltip`) to render the full content which is passed as the first argument */
   children?: (self: JSX.Element, isOpen?: boolean) => any;
   /** this is the full string */
-  label?: string | number;
+  label: string | number;
```
giogonzo: it doesn't make any sense to overflow with an empty tooltip right?

src/Scroll/ScrollView.tsx
```diff
@@ -17,9 +17,15 @@ export type ScrollViewDefaultProps = {
...
+export type ScrollTo = (
+  _x?: number | null,
+  _y?: number | null,
+  scrollDuration?: number
+) => void;
+
 export type ScrollViewRequiredProps = {
   /** components/nodes content. If you need to scroll programmatically pass a function and save `scrollTo(x, y, milliseconds)` callback for later use (it will be passed as first argument) ex: `(scrollTo) => { this.scrollTo = scrollTo; return <MyScrollViewContent />; }` */
-  children: React.ReactNode;
+  children: Children | ((scrollTo: ScrollTo) => Children);
```
giogonzo: explicitly also support the render-prop pattern (already supported at runtime)

src/Popover/Popover.tsx
```diff
@@ -60,7 +60,7 @@ export namespace Popover {
   export type Delay = number | { whenClosed?: number; whenOpen?: number };
 
   export type PopoverSettings = {
-    content: React.ReactNode;
+    content: NonNullable<Children>;
```
giogonzo: it doesn't make any sense to render an empty tooltip right? (this still allows for `false` to be passed here... anyway)

src/FocusableView/FocusableView.tsx
```diff
@@ -1,14 +1,11 @@
...
-import { props, t, ReactChildren } from "../utils";
+import { props, t, ReactChildren, Children } from "../utils";
 
 export type FocusableViewRequiredProps = {
   /** FocusableView content. If a function it gets called with the boolean "focused" */
-  children:
-    | ((focused: boolean) => React.ReactNode | React.ReactNode[])
-    | React.ReactNode
-    | React.ReactNode[];
+  children: ((focused: boolean) => Children) | Children;
```
giogonzo: `ReactNode[]` was superfluous (already part of `ReactNode` itself, it's recursive). Same goes for `Children`

src/TextOverflow/TextOverflow.tsx
gabro: I think this is a co-product masked by product. Both `children` and `label` are optional, but you can really only pass one of the two

giogonzo: good catch, makes sense, I'd probably update the typings by disallowing `children` then (or the other way around)?

gabro: can we make it a coproduct?

src/utils/index.ts
```diff
@@ -59,6 +59,6 @@ export const findDOMNode = <E extends Element = Element>(
 type ReactText = string | number;
 type _ReactChild = React.ReactElement<unknown> | ReactText;
 
-interface ChildrenArray extends Array<Children> {}
-type ReactFragment = ChildrenArray;
+export interface ChildrenArray extends Array<Children> {}
+export type ReactFragment = ChildrenArray;
```
gabro: why these exports?

giogonzo: TS is otherwise complaining with:
Screenshot 2019-09-17 at 19 58 27
I didn't really investigate this thoroughly 🤷‍♂

giogonzo: actually, it goes away if I use the `private` modifier for this method

giogonzo: but I can't just use the same "fix" for `render()` here:

Screenshot 2019-09-17 at 20 47 22

So I'd leave it as is now

PR #1414, opened 2020-01-02
#1360: proposal: remove runtime prop types checking (closes #1360)

### LINKED ISSUE #1360: proposal: remove runtime prop types checking
With all of our project now using TypeScript, and TS being quite adopted in the React community in general, we could remove prop types via tcomb `props` in order to:

1. spare some code and a runtime dependency
2. avoid cases where the two typings (runtime and static) differ in a subtle way

This should be done on a few other separate repos I think, such as `react-flexview`

giogonzo:
Closes #1360

As expected, the only interesting cases are the one previously involving tcomb `refinement`s. In a single case, it was possible to encode it at type-level (`TabbedPanel` tabs should be non-empty). For all the other cases (TimePicker, Meter mainly) I didn't invest time in expressing the invariants at type level, only reused the same checks performed at runtime inside of `componentDidMount`, and warning in case of issues.

## Test Plan

All checks are passing. This PR should have no impact at runtime, and I decided to not mark it as breaking for this reason. Let me know your thoughts

### tests performed

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -1,11 +1,11 @@
 import * as React from "react";
 import * as cx from "classnames";
 import every = require("lodash/every");
-import { props, t, stateClassUtil } from "../utils";
-import { ReactChild } from "tcomb-react";
+import { stateClassUtil } from "../utils";
 import { TextOverflow } from "../TextOverflow/TextOverflow";
 import FlexView from "react-flexview";
 import { LoadingSpinner } from "../LoadingSpinner/LoadingSpinner";
+import { warn } from "../utils/log";
 
 export type ButtonRequiredProps = {
   /** callback */
@@ -65,28 +65,8 @@ export namespace Button {
 
 type ButtonDefaultedProps = ButtonRequiredProps & ButtonDefaultProps;
 
-// types
-export const buttonStates: Button.ButtonState[] = [
-  "ready",
-  "not-allowed",
-  "processing",
-  "error",
-  "success"
-];
-const ButtonState = t.enums.of(buttonStates, "ButtonState");
-export const buttonTypes: Button.ButtonType[] = [
-  "default",
-  "primary",
-  "positive",
-  "negative",
-  "flat"
-];
-const ButtonType = t.enums.of(buttonTypes, "ButtonType");
-export const buttonSizes: Button.ButtonSize[] = ["tiny", "small", "medium"];
-const ButtonSize = t.enums.of(buttonSizes, "ButtonSize");
-
 // util
-const notBoth = (a: any, b: any): boolean => !(a && b);
+const notBoth = (a: unknown, b: unknown): boolean => !(a && b);
 const satisfyAll = (...conditions: Array<(props: Button.Props) => boolean>) => (
   props: ButtonRequiredProps
 ) => every(conditions, c => c(props));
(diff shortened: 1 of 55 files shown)


src/Button/Button.tsx
```diff
@@ -170,6 +126,14 @@ export class Button extends React.PureComponent<Button.Props> {
     style: {}
   };
 
+  componentDidMount() {
+    if (process.env.NODE_ENV !== "production") {
+      if (!satisfyAll(...propsInvariants)(this.props)) {
```
veej: is this spread useful?

src/FocusableView/FocusableView.tsx
```diff
@@ -133,7 +121,7 @@ export class FocusableView extends React.Component<FocusableView.Props> {
     return React.createElement(
       component as any,
       locals,
-      t.Function.is(children) ? children(focused) : children
+      isFunction(children) ? children(focused) : children
```
veej: why not `typeof children === "function"` as you did for strings and numbers?
That's the same exact thing lodash's `isFunction` does.

src/TimePicker/TimePicker.tsx
```diff
@@ -58,28 +53,19 @@ const getComponents = (
...
-    (!value || (lteTime(value, maxTime) && lteTime(minTime, value))),
-  "Props"
-);
+function validProps({ value, minTime, maxTime }: TimePicker.Props): boolean {
+  const validTimes =
+    (value == null || isTime(value)) &&
+    (minTime == null || isTime(minTime)) &&
+    (maxTime == null || isTime(maxTime));
+  const minMaxCoherent =
+    minTime != null && maxTime != null && lteTime(minTime, maxTime);
```
veej: Shouldn't this be `minTime == null || maxTime == null || lteTime(minTime, maxTime)`?

src/Button/Button.tsx
giogonzo: not sure, I left it as I found it in the code previously

src/FocusableView/FocusableView.tsx
giogonzo: makes sense, I'll do it

src/TimePicker/TimePicker.tsx
giogonzo: yup, good catch, thanks

src/Scroll/ScrollView.tsx
```diff
@@ -223,7 +226,7 @@ export class ScrollView extends React.Component<ScrollView.Props> {
           this.scrollView = sv;
         }}
       >
-        {t.Function.is(children) ? children(this.scrollTo) : children}
+        {isFunction(children) ? children(this.scrollTo) : children}
```
veej: replace with `typeof children === "function"`?

src/Tablo/Cell/Cell.tsx
```diff
@@ -101,15 +85,11 @@ export function _Cell<T extends {}, K extends keyof T | never = never>(
           vAlignContent={vAlignContent}
           hAlignContent={hAlignContent}
         >
-          {t.Function.is(render) ? render(data, rowData, rowIndex) : render}
+          {isFunction(render) ? render(data, rowData, rowIndex) : render}
```
veej: replace with `typeof render === "function"`?

src/Scroll/ScrollView.tsx
giogonzo: This change fails with a weird error:
```ts
This expression is not callable.
  Not all constituents of type '((scrollTo: ScrollTo) => Children) | (string & {}) | (number & {}) | (false & {}) | (true & {}) | (((scrollTo: ScrollTo) => Children) & string) | (((scrollTo: ScrollTo) => Children) & number) | ... 4 more ... | (((scrollTo: ScrollTo) => Children) & ReactPortal)' are callable.
    Type 'string & {}' has no call signatures.ts(2349)
```

Not sure how to fix without using an `any` cast really..

PR #1426, opened 2020-03-24
#1423: DatePicker should have clearable option (closes #1423)

### LINKED ISSUE #1423: DatePicker should have clearable option
At the moment, it's always clearable. It should give the option to be or not, like the dropdown.

There is a simple workaround, but the data type can still be null:

```
.SingleDatePickerInput .SingleDatePickerInput_clearDate {
          display: none;
}
```

aSlug:
Closes #1423

### tests performed

if `isClearable={false}` or not defined

- `value` has to be defined and it can't be `null`
- if defined, `onChange` has to be `(date: LocalDate) => void`
- there is no _clear_ icon

if `isClearable={true}`

- `value` has to be defined, but can be `null`
- if defined, `onChange` has to be `(date: LocalDate | null) => void`
- there is a _clear_ icon
----------

⚠️  **this is a breaking change** since the component was clearable by default and `value` was optional, now it's not clearable by default and `value` is always mandatory

### DIFF
--- a/src/DatePicker/DatePicker.tsx
+++ b/src/DatePicker/DatePicker.tsx
@@ -53,9 +53,9 @@ export namespace DatePicker {
         /** whether the user can remove the selected date */
         isClearable: true;
         /** current date */
-        value?: LocalDate;
+        value: LocalDate | null;
         /** called when value changes */
-        onChange?: (date?: LocalDate) => void;
+        onChange?: (date: LocalDate | null) => void;
       }
     | {
         /* whether the user can remove the selected date */
@@ -68,14 +68,14 @@ export namespace DatePicker {
 }
 
 export type State = {
-  value?: moment.Moment;
+  value: moment.Moment | null;
   hoveredDay?: moment.Moment;
   focused: boolean;
 };
 
 const valueToMomentDate: (
-  value?: LocalDate
-) => moment.Moment | undefined = value => (!value ? undefined : moment(value));
+  value?: LocalDate | null
+) => moment.Moment | null = value => (!value ? null : moment(value));
 
 const angleLeftIcon = (
   <svg width="10" height="10" viewBox="0 0 32 32">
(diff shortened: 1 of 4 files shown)


src/DatePicker/DatePicker.tsx
```diff
@@ -52,7 +48,23 @@ export namespace DatePicker {
     className?: string;
     /** an optional style object to pass to top level element of the component */
     style?: React.CSSProperties;
-  };
+  } & (
+    | {
+        /** whether the user can remove the selected date */
+        isClearable: true;
+        /** current date */
+        value?: LocalDate;
```
veej: I'd prefer a `LocalDate | null` here, instead of `undefined`. This would be consistent with other components where the "empty" value is represented with `null`.

src/FormField/DatePickerField.tsx
```diff
@@ -64,7 +55,7 @@ export class DatePickerField extends React.PureComponent<InternalProps> {
         hint={hint}
         render={(onFocus, onBlur) =>
           datePickerRenderer({
-            ...datePickerProps,
+            ...this.props,
```
veej: beware that here you're passing a lot of additional/unwanted props to the `DatePicker` underneath. Not sure this is something we want (and yes, spreading the DatePicker props at the top level wasn't really a good idea... we should change this for every field 🤔)

aSlug: True, but I had to do it because spreading and re-packing `this.props` was making it loose its conditional typing making it not usable as argument of `datePickerRenderer`. It's the easiest way and I think that now it creates no problem.
Otherwise if this is not acceptable I will find a way to extract `DatePicker.Props` from `DatePickerField.Props` preserving the typing

veej: oh god, you're right 🤦‍♂ 
Let's keep it like this atm. Could you please open a breaking issue to pack all these props inside a single `componentProps` in every field, instead of spreading them at top level?

aSlug:
@veej I applied the requested changes 👍
test plan has been updated

PR #1428, opened 2020-03-27
#1427: FormField components should have a prop packing all inner component's props instead of spreading them (closes #1427) [WIP]

### LINKED ISSUE #1427: FormField components should have a prop packing all inner component's props instead of spreading them
Now the props of the components in the FormField module are something like this

```
type XFieldProps = {
  ... lot of props ...
} & X.Props;
```
this leads to complications when managing the props from `X.Props`, it would be better to pack them all in a single prop, like
```
type XProps = {
  ... lot of props ...,
  innerXProps: X.Props
}
```

aSlug:
Closes #1427

🚧 WIP  🚧


aSlug:
Right, I will update the issue accordingly 👍

PR #1429, opened 2020-03-27
#1427: FormField components should have a prop packing all inner component's props instead of spreading them (closes #1427)

### LINKED ISSUE #1427: FormField components should have a prop packing all inner component's props instead of spreading them

aSlug:
Closes #1427

### tests performed

Schermata 2020-03-27 alle 15 59 03

All the components in the FormField module now have a mandatory property requiring an object containing all the properties of their inner component. This object has the same type of the `Props` of the inner component.

No change in the UX has been introduced

### DIFF
--- a/src/FormField/InputField.tsx
+++ b/src/FormField/InputField.tsx
@@ -21,7 +21,7 @@ type NonDefaultProps = {
   className?: string;
   /** an optional style object to pass to top level element of the component */
   style?: React.CSSProperties;
-  /** the properties fo the input */
+  /** the properties of the input */
   inputProps: Input.Props;
 };
 
--- a/src/FormField/InputField.tsx
+++ b/src/FormField/InputField.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import { ObjectOmit } from "../utils";
 import * as cx from "classnames";
 import Input from "../Input";
 import { FormField } from "./FormField";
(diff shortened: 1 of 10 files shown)


src/FormField/PasswordInputField.tsx
```diff
@@ -22,11 +21,10 @@ type FieldProps = {
   className?: string;
   /** an optional style object to pass to top level element of the component */
   style?: React.CSSProperties;
+  /** the properties of password input */
+  passwordInputProps: PasswordInput.Props;
 };
 
-type NonDefaultProps = FieldProps &
-  ObjectOmit<PasswordInput.Props, keyof FieldProps>;
-
```
aSlug: could you double-check if removing this and using all the properties of `PasswordInput.Props` is legit? I found no conflict but I'm not 100% confident that I'm not missing anything

src/FormField/InputField.tsx
```diff
@@ -9,22 +8,23 @@ type DefaultProps = {
...
   required?: FormField.Props["required"];
   /** optional props to pass to the wrapping View */
   viewProps?: FormField.Props["viewProps"];
+  /** an optional hint describing what's the expected value for the field (e.g. sample value or short description) */
   hint?: FormField.Props["hint"];
   /** an optional class name to pass to top level element of the component */
   className?: string;
   /** an optional style object to pass to top level element of the component */
   style?: React.CSSProperties;
+  /** the properties fo the input */
```
veej: typo: `fo` => `for`

src/FormField/PasswordInputField.tsx
veej: looks legit 👍

PR #1437, opened 2020-04-10
#1432: Showroom examples for InputField is broken (closes #1432)

### LINKED ISSUE #1432: Showroom examples for InputField is broken
See: https://react-components.buildo.io/#formfield

Seems to be a regression introduced by https://github.com/buildo/react-components/pull/1429

aSlug:
Closes #1432

### tests performed

`yarn start` and verified that no example is broken


aSlug:
@giogonzo yes the `xxxProps` prop is mandatory in every `xxxForm` component, but since `value` and `onChange` are always mandatory It will never be empty

PR #1438, opened 2020-04-10
#1436: Add parameter to customize anchor of Menu component (closes #1436)

### LINKED ISSUE #1436: Add parameter to customize anchor of Menu component
The popover used inside Menu has an hardcoded prop `anchor = "end"`. This is ok if the menu is used on the right side of the view, but if the user wants to use it on the left side it has to force its anchoring editing the css.

It would be useful if `Menu` had a `anchor` property used in the same way of `position`

aSlug:
Closes #1436

### tests performed

The horizontal alignment of the Menu is customizable using the `anchor` prop.

Position is "bottom" and anchor is "end" by default, to avoid introduce breaking changes. This also fixes an error in the examples.

`anchor="start"`           |  `anchor="end"` or not defined
:-------------------------:|:-------------------------:
Schermata 2020-04-10 alle 16 59 52  |  Schermata 2020-04-10 alle 16 59 30

### DIFF
--- a/src/Menu/Examples.md
+++ b/src/Menu/Examples.md
@@ -4,29 +4,29 @@
 intialState = { isOpen: false };
 
 function toggleOpen() {
-  setState({ isOpen: !state.isOpen })
+  setState({ isOpen: !state.isOpen });
 }
 
 function onClick(option) {
-  alert(`${option.title} clicked`)
+  alert(`${option.title} clicked`);
 }
 
 const { isOpen } = state;
 
 const options = [
   {
-    title: 'Preferences',
-    type: 'item',
+    title: "Preferences",
+    type: "item",
     onClick
   },
   {
-    title: 'Change password',
-    type: 'item',
+    title: "Change password",
+    type: "item",
     onClick
   },
   {
-    title: 'Logout',
-    type: 'item',
+    title: "Logout",
+    type: "item",
     onClick
   }
 ];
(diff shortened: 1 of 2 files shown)


src/Menu/Examples.md
```diff
@@ -4,29 +4,29 @@
 intialState = { isOpen: false };
 
 function toggleOpen() {
-  setState({ isOpen: !state.isOpen })
+  setState({ isOpen: !state.isOpen });
```
veej, line 7: are these changes ok? are we formatting the other examples?

src/Menu/Menu.tsx
```diff
@@ -27,8 +27,10 @@ export type MenuDefaultProps = {
   isOpen: boolean;
   /** whether the menu should be closed when clicking outside */
   dismissOnClickOutside: boolean;
-  /** wheter the menu should be rendered on top or at the bottom of the trigger */
+  /** wheter the menu should be rendered on top or at the bottom of the trigger (bottom by default) */
   position: "top" | "bottom";
+  /** wheter the menu should be aligned with the start, the end or the center of the trigger (end by default) */
```
veej, line 32: typo: Whether

src/Menu/Menu.tsx
```diff
@@ -27,8 +27,10 @@ export type MenuDefaultProps = {
...
+  anchor: "start" | "center" | "end";
```
veej, line 33: could it make sense to take these props from the Popover props?
```
anchor: Popover.Props['anchor']
```

same for `position` above.

aSlug: Sure but `Popover.Props['position']` can also be "left" or "right", I don't know if we want to make it so general. 
Actually for the same reason i initially didn't include "center" as an option for `anchor`.

src/Menu/Examples.md
aSlug: actually i didn't notice this, but looks compliant with the standard formatting we use everywhere.
Didn't it?

veej: Examples are using inconsistent formatting. I'm ok with it.

src/Menu/Menu.tsx
veej: 👍 makes sense

src/Menu/Examples.md
aSlug: Yes i just verified and indeed there is an overall inconsistency in examples, but I think this should be discussed in a separate issue.

PR #1440, opened 2020-04-17
#1439: fix overall inconsistency of examples (closes #1439)

### LINKED ISSUE #1439: fix overall inconsistency of examples
The snippets of code in the `examples.md` files are not formatted following the same rules.

aSlug:
Closes #1439

I changed the name of `.prettierrc.yml` since it was missing the initial `.` so it was not actually being used, then i ran `yarn prettier-fix` to apply the formatting rules to the whole project.


aSlug:
I didn't notice that! Indeed it's quite strange... I will investigate about it 👍

giogonzo:
@aSlug I verified it: if I pull master and edit an Example file, "save without format", then run prettier-fix -> nothing happens (and `prettier-check` does not complain).
I think it's sufficient to adapt the package scripts to also consider .md files in order to enforce this CI-side

aSlug:
Yes I just realized that I manually ran it on every `example.md` file, intentionally avoiding `README.md` files, and after that i ran `yarn prettier-fix`

PR #1474, opened 2021-09-06
Fix compile issue on yarn 1

### LINKED ISSUE #1459: Allow non-string headers in TabbedPanel
Currently TabbedPanel accepts only string headers. However, since it simply renders the header, accepting other React elements should be possible, and it would be useful to display SVG icons.

As an example, passing a header like
```jsx
(
  <View shrink={false} vAlignContent="bottom">
    <MySvgIcon/>
    {label}
  </View>
) as any,
```
(bypassing type checking) seems to work OK.

veej:
### tests performed

The project is compiling fine

### DIFF
--- a/src/DatePicker/datePicker.scss
+++ b/src/DatePicker/datePicker.scss
@@ -1,5 +1,5 @@
 @import '../utils/theme/variables.scss';
-@import 'react-dates/lib/css/_datepicker';
+@import '~react-dates/lib/css/_datepicker.css';
 
 $date-picker-width-input: 150px !default;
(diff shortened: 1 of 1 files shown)


src/DatePicker/datePicker.scss
```diff
@@ -1,5 +1,5 @@
 @import '../utils/theme/variables.scss';
-@import 'react-dates/lib/css/_datepicker';
```
veej, line 2: unrelated. Fixes the fix applied [here](https://github.com/buildo/react-components/pull/1471) as it seems to not work with yarn 1

PR #1477, opened 2021-09-24
1476 radio group doesnt support custom values

### LINKED ISSUE #1476: RadioGroup doesn't support custom values
RadioGroup supports only options with string values.
It should be a generic component instead, supporting values of any type.

veej:
Closes #1476 

### Test performed
image

### DIFF
--- a/src/DatePicker/datePicker.scss
+++ b/src/DatePicker/datePicker.scss
@@ -1,5 +1,5 @@
 @import '../utils/theme/variables.scss';
-@import 'react-dates/lib/css/_datepicker';
+@import '~react-dates/lib/css/_datepicker.css';
 
 $date-picker-width-input: 150px !default;
 $date-picker-month-header-background: $brc-azure !default;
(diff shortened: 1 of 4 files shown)


src/DatePicker/datePicker.scss
```diff
@@ -1,5 +1,5 @@
 @import '../utils/theme/variables.scss';
-@import 'react-dates/lib/css/_datepicker';
+@import '~react-dates/lib/css/_datepicker.css';
```
giogonzo, line 2: related?

src/RadioGroup/RadioGroup.tsx
```diff
@@ -2,18 +2,19 @@ import * as React from 'react';
 import * as cx from 'classnames';
 import FlexView from 'react-flexview';
 
-export type RadioOption = {
+export type RadioOption<T = string> = {
```
giogonzo, line 5: do we need the `= default` at all?

src/DatePicker/datePicker.scss
veej: I had to rebase on https://github.com/buildo/react-components/pull/1474 (not yet merged) to make it compile.

src/RadioGroup/RadioGroup.tsx
veej: maybe not

giogonzo (approved):

PR #1489, opened 2022-05-10
update scss to support sass lib

neomusic:
These PR serves to support new versions of Sass. In reality the bar as a division is deprecated and will soon be eliminated.
Check https://sass-lang.com/documentation/breaking-changes/slash-div


src/LoadingSpinner/loadingSpinner.scss
```diff
@@ -26,8 +26,8 @@ $thickness: 0.1em;
       -ms-transform: translateZ(0);
       -webkit-transform: translateZ(0);
       transform: translateZ(0);
-      margin-top: -$size / 2;
-      margin-left: -$size / 2;
+      margin-top: - calc($size / 2);
```
veej, line 29: just to be sure, is it working correctly? The minus before the `calc` and with a space in-between looks weird

neomusic: After some test this solution is not perfect. Close PR

PR #1495, opened 2026-04-21
Forward data-* attributes on Button, DatePicker, and Dropdown variants

MalpenZibo:
## Summary
- Add a `DataAttributes` mixin type and `pickDataAttributes` runtime helper in `src/utils` that extracts `data-*` keys from a props object.
- Forward `data-*` attributes on the root element of `Button`, `DatePicker`, `Dropdown`, `MultiDropdown`, and `MultiDropdownWithSelectAll` so consumers can set `data-testid` (and any other `data-*`) directly instead of wrapping these components.
- For the `Dropdown` variants, since `react-select` does not forward unknown props, wrap the select in a `display: contents` `<div>` only when `data-*` are present — the DOM is otherwise unchanged.

## Test plan
- [ ] Pass `data-testid` to `Button`, `StatefulButton`, `DatePicker`, `Dropdown`, `MultiDropdown`, and `MultiDropdownWithSelectAll` and confirm the attribute lands on the rendered root element.
- [ ] Verify that when no `data-*` props are passed, the rendered DOM is unchanged (no extra wrapper on the Dropdown variants).
- [ ] Type-check the workspace (`yarn tsc`) to confirm `DataAttributes` integrates without excess-property regressions on existing callers.

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import cx from 'classnames';
 import every = require('lodash/every');
-import { stateClassUtil } from '../utils';
+import { DataAttributes, pickDataAttributes, stateClassUtil } from '../utils';
 import { TextOverflow } from '../TextOverflow/TextOverflow';
 import FlexView from 'react-flexview';
 import { LoadingSpinner } from '../LoadingSpinner/LoadingSpinner';
@@ -48,7 +48,7 @@ export namespace Button {
   export type ButtonType = 'default' | 'primary' | 'positive' | 'negative' | 'flat';
   export type ButtonSize = 'tiny' | 'small' | 'medium';
 
-  export type Props = ButtonRequiredProps & Partial<ButtonDefaultProps>;
+  export type Props = ButtonRequiredProps & Partial<ButtonDefaultProps> & DataAttributes;
 }
 
 type ButtonDefaultedProps = ButtonRequiredProps & ButtonDefaultProps;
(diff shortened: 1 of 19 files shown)


src/Button/Button.tsx
```diff
@@ -194,7 +194,7 @@ export class Button extends React.PureComponent<Button.Props> {
     const loading = buttonState === 'processing';
 
     return (
-      <div className="button" style={wrapperStyle}>
+      <div className="button" style={wrapperStyle} {...pickDataAttributes(this.props)}>
```
pedrodim: Aggiungiamo direttamente `role="button"` o troppo out-of-scope per questa PR? 😄

src/Dropdown/commons.tsx
```diff
@@ -58,3 +60,25 @@ export const getCommonClassnames = (
 export const defaultStyle = {
   input: () => ({ margin: 0, padding: 0 })
 };
+
+/**
+ * react-select does not forward unknown props (including data-*) to its root
+ * element. When the consumer passes any data-* attribute, wrap the select in a
```
pedrodim: Usando un [custom](https://github.com/buildo/react-components/blob/master/src/Dropdown/commons.tsx#L37C14-L42) `SelectContainer`, che fa da wrapper a quello vero di react-select, possiamo aggiungere il supporto per `data-*` senza il div extra.

Non avrai accesso alle prop, ma possiamo saltare questo con un context per passare tutte le prop senza dipendere da cosa fa internamente la libreria (Telerik per esempio ha un `propContext` proprio per questo).

pedrodim (approved):

